PR #7, opened 2018-01-27
Implement random access to index contents

charles-dyfis-net:
This is a building block on which nbd, mount, seek-capable caidx parsing and similar extensions can be built.

Needs test coverage; consider this proof-of-concept-level code.

### DIFF
--- a/cmd/desync/cat.go
+++ b/cmd/desync/cat.go
@@ -1,0 +1,135 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"net/url"
+	"os"
+
+	"github.com/folbricht/desync"
+	"io"
+)
+
+const catUsage = `desync cat [options] <caibx>
+
+Stream a caibx to stdout, optionally seeking and limiting the read length.
+
+This is inherently slower than extract as while multiple chunks can be retrieved
+concurrently, writing to stdout cannot be parallelized.`
+
+func cat(ctx context.Context, args []string) error {
+	var (
+		cacheLocation  string
+		n              int
+		err            error
+		storeLocations = new(multiArg)
+		stores         []desync.Store
+		offset         int
+		length         int
+		readIndex      bool
+	)
+	flags := flag.NewFlagSet("cat", flag.ExitOnError)
+	flags.Usage = func() {
+		fmt.Fprintln(os.Stderr, catUsage)
+		flags.PrintDefaults()
+	}
+
+	flags.BoolVar(&readIndex, "i", false, "Read index file (caidx), not catar, in 2-argument mode")
+	flags.Var(storeLocations, "s", "casync store location, can be multiples")
+	flags.StringVar(&cacheLocation, "c", "", "use local store as cache")
+	flags.IntVar(&n, "n", 10, "number of goroutines")
+	flags.IntVar(&offset, "o", 0, "offset in bytes to seek to before reading")
+	flags.IntVar(&length, "l", 0, "number of bytes to read")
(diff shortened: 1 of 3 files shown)


readseeker.go
```diff
@@ -0,0 +1,153 @@
...
+func (ip *IndexPos) Seek(offset int64, whence int) (int64, error) {
+	var newPos int64
+	var err error
+	switch whence {
+	case io.SeekStart:
+		newPos = offset
+	case io.SeekCurrent:
+		newPos = ip.pos + offset
+	case io.SeekEnd:
+		newPos = ip.Length - offset
```
folbricht: I believe this should be:
```go
newPos = ip.Length + offset
```
but since that doesn't really work for this use case, perhaps simply return EOF

readseeker.go
```diff
@@ -0,0 +1,153 @@
...
+			}
+		}
+		chunkRemainingBytes := ip.curChunk[ip.curChunkOffset:len(ip.curChunk)]
+		if len(chunkRemainingBytes) == 0 && ip.curChunkIdx == (len(ip.Index.Chunks)-1) {
+			break // if running into the end after successful read, return a short read
+		}
+		copiedBytes := copy(remainingBytes, chunkRemainingBytes)
+		remainingBytes = remainingBytes[copiedBytes:]
+		totalCopiedBytes += copiedBytes
+		_, err := ip.Seek(int64(copiedBytes), io.SeekCurrent)
```
folbricht: Pretty sure this `err` masks the one returned at the end of the function. So an error by Seek() isn't going to be returned. Removing `:` should do the trick.

readseeker.go
```diff
@@ -0,0 +1,153 @@
...
+	Length int64 // total length of file
+	pos    int64 // Location within offset stream; must be 0 <= Pos <= Index.
+
+	curChunkID     ChunkID // hash of current chunk
+	curChunk       []byte  // decompressed version of current chunk
+	curChunkIdx    int     // identity of current chunk
+	curChunkOffset int64   // offset within current chunk
+}
+
+func NewIndexReadSeeker(i Index, s Store) IndexPos {
```
folbricht: Given the nature of this reader, may want to consider returning *IndexPos instead

cmd/desync/cat.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// Read the input
+	c, err := readCaibxFile(inFile)
+	if err != nil {
+		return err
+	}
+
+	// Write the output
+	readSeeker := desync.NewIndexReadSeeker(c, s)
+	readSeeker.Seek(int64(offset), io.SeekStart)
```
folbricht: May want to add error handling for this
```go
if err = readSeeker.Seek(int64(offset), io.SeekStart); err != nil {
  return err
}
```

cmd/desync/cat.go
```diff
@@ -0,0 +1,135 @@
...
+	"flag"
+	"fmt"
+	"net/url"
+	"os"
+
+	"github.com/folbricht/desync"
+	"io"
+)
+
+const catUsage = `desync cat [options] <caibx>
```
folbricht: This one allows an optional output file, should it say `desync cat [options] <caibx> [<outputfile>]` ?

readseeker.go
charles-dyfis-net: This is a place where I misread the interface we were implementing early on, and didn't catch a consequence of that when fixing it. Thought offset was uint64, and thus that adding would make no sense; since it's int64, we can move backwards, so `+` is indeed correct.

PR #15, opened 2018-03-12
Feature/tls client auth

aspeteRakete:
Adds the ability to define certificate/key to do TLS Client Authentication.

### DIFF
--- a/cmd/desync/cache.go
+++ b/cmd/desync/cache.go
@@ -62,7 +62,7 @@ func cache(ctx context.Context, args []string) error {
 	}
 
 	// Parse the store locations, open the stores and add a cache is requested
-	s, err := MultiStoreWithCache(n, cacheLocation, storeLocations.list...)
+	s, err := MultiStoreWithCache(n, cacheLocation, "", "",storeLocations.list...)
 	if err != nil {
 		return err
 	}
--- a/cmd/desync/cache.go
+++ b/cmd/desync/cache.go
@@ -21,6 +21,8 @@ func cache(ctx context.Context, args []string) error {
 		cacheLocation  string
 		n              int
 		storeLocations = new(multiArg)
+		clientCert string
+		clientKey string
 	)
 	flags := flag.NewFlagSet("cache", flag.ExitOnError)
 	flags.Usage = func() {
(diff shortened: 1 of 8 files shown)


cmd/desync/cache.go
```diff
@@ -32,6 +34,8 @@ func cache(ctx context.Context, args []string) error {
 	flags.StringVar(&cacheLocation, "c", "", "use local store as cache")
 	flags.IntVar(&n, "n", 10, "number of goroutines")
 	flags.BoolVar(&desync.TrustInsecure, "i", false, "allow invalid certificates")
+	flags.StringVar(&clientCert, "clientCert", "", "Client Certificate for TLS authentication")
```
charles-dyfis-net: Might be helpful to specify that this is expected to be a filename. ("Filename with client certificate [...]"?); please take this as given for the similar instances of this change elsewhere.

cmd/desync/cache.go
```diff
@@ -62,7 +70,7 @@ func cache(ctx context.Context, args []string) error {
 	}
 
 	// Parse the store locations, open the stores and add a cache is requested
-	s, err := MultiStoreWithCache(n, cacheLocation, storeLocations.list...)
+	s, err := MultiStoreWithCache(n, cacheLocation, clientCert, clientKey,storeLocations.list...)
```
charles-dyfis-net: Minor stylistic quibble, but it'd be consistent to have a space after the last inserted comma.

cmd/desync/cat.go
```diff
@@ -32,6 +32,8 @@ func cat(ctx context.Context, args []string) error {
 		offset         int
 		length         int
 		readIndex      bool
+		clientCert string
```
charles-dyfis-net: Minor stylistic quibble, but it'd be good for the type name (`string`) to line up with the `bool` from the lines above.

remotehttp.go
```diff
@@ -34,13 +35,37 @@ func NewRemoteHTTPStore(location *url.URL, n int) (*RemoteHTTP, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to create CaCertPool")
+		}
+		tr = &http.Transport{
+			DisableCompression:  true,
+			MaxIdleConnsPerHost: n,
+			TLSClientConfig:     &tls.Config{
+				InsecureSkipVerify: TrustInsecure,
+				Certificates: []tls.Certificate{certificate},
+				RootCAs:      caCertPool,},
```
charles-dyfis-net: If putting the `}` on the same line as the last entry, I'd argue that that defeats the value of the trailing comma (such being permitting new key/value pairs to be added without changing any other lines).

aspeteRakete:
Hi. Thanks for the comments. go fmt'ed everything and added the Path hint to the flag description.

folbricht:
Thank you for the contribution.

PR #17, opened 2018-03-14
return pErr instead of nil

### LINKED ISSUE #16: desync.Touch doesn't return error when chunks missing
I just tried to Populate a local store/cache with Touch with a malformed store url. The Server Responds with a 404, the Process stops but the returned error is nil.

aspeteRakete:
Return pErr instead of nil in desync.Touch Fixes #16

### DIFF
--- a/touch.go
+++ b/touch.go
@@ -60,5 +60,5 @@ loop:
 	close(in)
 	wg.Wait()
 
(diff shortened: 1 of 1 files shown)


folbricht:
Nice catch, thanks

PR #40, opened 2018-07-27
Add support for AWS temporary (refreshable) credentials

angelcar:
## What is being added?
We need to access S3 stores using [aws temporary credentials ](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html) as opposed to static  access and secret keys. Temp credentials are a more secure way of handling access to S3 since credentials are valid for a short window of time and they need to be refreshed periodically by other services (i.e. [AWS STS assume role](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) or [AWS SSM](https://docs.aws.amazon.com/systems-manager/latest/userguide/automation-permissions.html)).

Temp credentials are read from a [standard aws credentials file](https://docs.aws.amazon.com/cli/latest/userguide/cli-config-files.html) which is specified desync's config file. 

This change is backwards compatible, so the users now have the option to specify their credentials directly in the config file or point to a specific profile inside the aws creds file. If a profile is not specified, the [default] profile is used.

Additionally, users can now specify an aws region explicitly. Providing the region explicitly to minio improves its performance, according to [minio's documentation](https://github.com/minio/minio-go/blob/b72bb4bde10032d4e9d3e3175b79ddcdc3efdd85/api.go#L174). If the region is not specified, an empty string is passed downstream (current behavior).

Lastly, the precedence for credential resolution is: first, env variables, then access-key/secret-key in the config file, and aws-credentials-file last. 

example:

~/.config/desync/config.json
```
{
  "http-timeout": 60000000000,
  "http-error-retry": 0,
  "s3-credentials": {
       "http://localhost": {
           "access-key": "MYACCESSKEY",
           "secret-key": "MYSECRETKEY"
       },
       "https://127.0.0.1:9000": {
           "aws-credentials-file": "/Users/user/.aws/credentials",
       },
       "https://127.0.0.1:8000": {
           "aws-credentials-file": "/Users/user/.aws/credentials",
           "aws-profile": "profile_static"
       },
       "https://s3.us-west-2.amazonaws.com": {
           "aws-credentials-file": "/Users/user/.aws/credentials",
           "aws-region": "us-west-2",
           "aws-profile": "profile_refreshable"
       }
  }
}
```

~/.aws/credentials
```
[default]
aws_access_key_id = DEFAULT_PROFILE_KEY
aws_secret_access_key = DEFAULT_PROFILE_SECRET

[profile_static]
aws_access_key_id = OTHERACCESSKEY
aws_secret_access_key = OTHERSECRETKEY

[profile_refreshable]
aws_access_key_id = PROFILE_REFRESHABLE_KEY
aws_secret_access_key = PROFILE_REFRESHABLE_SECRET
aws_session_token = PROFILE_REFRESHABLE_TOKEN
```

## Testing done 
* Added unit tests for new code
* Tested for backwards compatibility 
* Tested new feature with refreshable credentials

## TODO
* When credentials are invalid, the error returned is that the chunk doesn't exist. Ideally we want to see that the chunk could not be retrieved due to credentials being invalid. Will submit that as a bug fix in a separate PR at some other point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ go get -u github.com/folbricht/desync/cmd/desync
 ### Environment variables
 - `CASYNC_SSH_PATH` overrides the default "ssh" with a command to run when connecting to a remote SSH or SFTP chunk store
 - `CASYNC_REMOTE_PATH` defines the command to run on the chunk store when using SSH, default "casync"
-- `S3_ACCESS_KEY` and `S3_SECRET_KEY` can be used to define S3 store credentials if only one store is used. Caution, these values take precedence over any S3 credentials set in the config file.
+- `S3_ACCESS_KEY`, `S3_SECRET_KEY`, `S3_REGION` can be used to define S3 store credentials if only one store is used. Caution, these values take precedence over any S3 credentials set in the config file.
 
 ### Caching
 The `-c <store>` option can be used to either specify an existing store to act as cache or to populate a new store. Whenever a chunk is requested, it is first looked up in the cache before routing the request to the next (possibly remote) store. Any chunks downloaded from the main stores are added to the cache. In addition, when a chunk is read from the cache and it is a local store, mtime of the chunk is updated to allow for basic garbage collection based on file age. The cache store is expected to be writable. If the cache contains an invalid chunk (checksum does not match the chunk ID), the operation will fail. Invalid chunks are not skipped or removed from the cache automatically. `verfiy -r` can be used to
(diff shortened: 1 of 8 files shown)


folbricht:
Thanks a lot for the contribution and the detailed explanation (I didn't know about temp creds).

As for the wrong error message when using invalid creds, I filed https://github.com/folbricht/desync/issues/43 for tracking. You're welcome to give that one a try as well.

Issue #37, opened 2018-05-24
chunk id X does not match its hash Y


limbo127:
Hello, 
we are using a lot of desync, and we have some times bad chunk .
desync make -m  "64:256:512" -n 4  -s /data/cache2/store /data/cache2/indexes/file.index /data/file.raw

sometimes we cannot use our images, and a desync verify show  some  chunk id X does not match ...
we can delete wrong chunk but images is lost.
How can we do ? 
- desync make option to verify chunk sha ?
- is it a possible disk issue ? ( ssd hard disk, f2fs partition )

Regards, 
Nicolas Prochazka

folbricht:
I don't think that desync would store a chunk incorrectly, unless it's maybe killed halfway through writing a chunk. Other than that I suspect that perhaps something else is modifying the file or as you said there's a filesystem/disk issue.

As for verifying, would a separate command work? Adding an option to every single command may be a bit messy. So I'm thinking of something along the lines of `desync verify -s <store> index.caibx`

limbo127:
there's no error during desync make process.
what is hapenning if two desync make process are working on the same local
store ?
regards

2018-05-24 17:25 GMT+02:00 folbricht <notifications@github.com>:

limbo127:
thank,
I 'm doing some test today
Regards
nicolas

2018-05-25 3:23 GMT+02:00 folbricht <notifications@github.com>:

limbo127:
desync make -m 64:256:512 -n 3 -s /data/cache2/store /data/c5df8a5cdd1e6e26fd7ce210c5bbf758.6e4cd4c9-5e8b-4ada-98ea-0b4643370eeb.index /data/c5df8a5cdd1e6e26fd7ce210c5bbf758.6e4cd4c9-5e8b-4ada-98ea-0b4643370eeb.raw



data to be stored does not match expected hash 91ba20c831c892fa32b6b7e8f2e2b3264bb4605b71c9269b77b5f88f8a5dddd5, got ??FA?@?????"??{?_????Q}b?B??

folbricht:
Thank you. This error suggests the issue is actually in desync. I'll need to review that code-path. Would you be able to provide more details of how you repro this?
- Does it always fail in the same place/same file?
- How large is the file and is there anything unusual about it (mostly 0 bytes or such)?
- Would you be able to provide an example file I could use that triggers the problem?

limbo127:
Does it always fail in the same place/same file?
[...]

PR #46, opened 2018-07-31
Encapsulate chunk storage in a single place

angelcar:
# Changes
This change is just a refactor to encapsulate chunk storage in a single place. Now `chop`, `make` and `tar` commands use the same piece of code to store chunks. The duplicated code being encapsulated was repeated in a couple of places and that already made a bug fix ineffective since it was applied to one place but not the other (copy paste was needed). For more info about the bug please see https://github.com/folbricht/desync/issues/37

In addition, I think chop.go was unnecessarily re-reading the data file [here](https://github.com/folbricht/desync/blob/master/chop.go#L59) since `IndexFromFile` was already chunking and reading the file [here](https://github.com/folbricht/desync/blob/master/make.go#L62). I considered this duplicated code. The only funny thing I had to do to make this work was to declare the `[]byte` directly in `IndexChunk` so that I can pass it around between functions. I realize that this might not be what you intended since `IndexChunk` contains only things that you are writing to the index file, but I figured that as long as `Index.WriteTo` ignores this new field, its all good.

Can you confirm I am not missing anything here?

Cheers :)

# Testing done

* desync unit tests passed
* Chopped and assembled back a sample file
* Executed tar/untar and make/extract with new files and the tool works as expected
* Tested for backwards compatibility by chopping a file using a binary with my changes in place and assembled back the file using a binary without my changes. This proves that my changes are backwards compatible.

### DIFF
--- a/chunker.go
+++ b/chunker.go
@@ -1,7 +1,6 @@
 package desync
 
 import (
-	"crypto/sha512"
 	"errors"
 	"fmt"
 	"io"
@@ -144,7 +143,7 @@ func (c *Chunker) fillBuffer() (n int, err error) {
 
 // Next returns the starting position as well as the chunk data. Returns
 // an empty byte slice when complete
-func (c *Chunker) Next() (Chunk, error) {
+func (c *Chunker) Next() (uint64, []byte, error) {
 	if len(c.buf) < int(c.max) {
 		n, err := c.fillBuffer()
 		if err != nil {
@@ -200,7 +199,7 @@ func (c *Chunker) Next() (Chunk, error) {
 	}
 }
 
-func (c *Chunker) split(i int, err error) (Chunk, error) {
+func (c *Chunker) split(i int, err error) (uint64, []byte, error) {
 	// save the remaining bytes (after the split position) for the next round
 	start := c.start
 	b := c.buf[:i]
@@ -210,8 +209,7 @@ func (c *Chunker) split(i int, err error) (Chunk, error) {
 	// reset the hash
 	c.hIdx = 0
 	c.hValue = 0
-	id := sha512.Sum512_256(b)
-	return Chunk{ID: id, Start: start, Data: b}, err
+	return start, b, err
 }
 
 // Min returns the minimum chunk size
--- a/chunker.go
+++ b/chunker.go
@@ -1,6 +1,7 @@
 package desync
 
 import (
+	"crypto/sha512"
 	"errors"
 	"fmt"
 	"io"
@@ -143,7 +144,7 @@ func (c *Chunker) fillBuffer() (n int, err error) {
 
 // Next returns the starting position as well as the chunk data. Returns
 // an empty byte slice when complete
-func (c *Chunker) Next() (uint64, []byte, error) {
+func (c *Chunker) Next() (Chunk, error) {
 	if len(c.buf) < int(c.max) {
 		n, err := c.fillBuffer()
 		if err != nil {
@@ -199,7 +200,7 @@ func (c *Chunker) Next() (uint64, []byte, error) {
 	}
 }
 
-func (c *Chunker) split(i int, err error) (uint64, []byte, error) {
+func (c *Chunker) split(i int, err error) (Chunk, error) {
 	// save the remaining bytes (after the split position) for the next round
 	start := c.start
 	b := c.buf[:i]
@@ -209,7 +210,8 @@ func (c *Chunker) split(i int, err error) (uint64, []byte, error) {
 	// reset the hash
 	c.hIdx = 0
 	c.hValue = 0
-	return start, b, err
+	id := sha512.Sum512_256(b)
+	return Chunk{ID: id, Start: start, Data: b}, err
 }
 
 // Min returns the minimum chunk size
--- a/chunkstorage.go
+++ b/chunkstorage.go
@@ -2,7 +2,6 @@ package desync
 
 import (
 	"bytes"
-	"context"
 	"fmt"
 	"os"
 	"sync"
@@ -12,12 +11,7 @@ import (
 
 type ChunkStorage struct {
 	sync.RWMutex
-	cancel context.CancelFunc
-	n      int
 	ws     WriteStore
-	pb     ProgressBar
-	wg     sync.WaitGroup
-	pErr   error
 	stored map[ChunkID]struct{}
 }
 
--- a/chunkstorage.go
+++ b/chunkstorage.go
@@ -111,7 +111,7 @@ func (s *ChunkStorage) markAsStored(c Chunk) {
 func (s *ChunkStorage) StoreChunk(c Chunk) error {
 
 	// Check in-memory cache to see if chunk has been stored, if so, skip it
-	if s.stored[c.ID] {
+	if s.isChunkStored(c) {
 		return nil
 	}
 
--- a/chunkstorage.go
+++ b/chunkstorage.go
@@ -11,119 +11,56 @@ import (
 )
 
 type ChunkStorage struct {
-	sync.Mutex
-	ctx    context.Context
+	sync.RWMutex
 	cancel context.CancelFunc
 	n      int
 	ws     WriteStore
-	in     <-chan Chunk
 	pb     ProgressBar
-
 	wg     sync.WaitGroup
 	pErr   error
-	stored map[ChunkID]bool
+	stored map[ChunkID]struct{}
 }
 
 // Stores chunks passed in the input channel asynchronously. Wait() will wait for until the input channel is closed or
 // until there's an error, in which case it will return it.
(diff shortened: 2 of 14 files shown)


angelcar:
Update: I think I just figured out why you were reading the file again. The hop command needs to read the file provided an index... oops!

Anyways, I fixed that and included chop to my tests, which I was totally neglecting. I re-tested everything as described before, and in addition I tested that the following works

```
desync make  -s ./test-store ./test.caidx ./blob
desync chop -s ./new-store ./test.caidx ./blob
desync extract  -s ./new-store ./test.caidx ./new-blob
shasum -a 256 ./blob
7e095525bcf72c31ac9f916002da9ac2494bc9ffb9b427118a96dceee5166c11
shasum -a 256 ./new-blob
7e095525bcf72c31ac9f916002da9ac2494bc9ffb9b427118a96dceee5166c11
``` 
Cheers,
Angel.

folbricht:
Thanks a lot for this PR, much appreciated. Seems like I procrastinated long enough for someone else to take on some of the code-duplication :) I haven't had the time yet to review it in detail but here's a few comments so far:
1) Sorry you had to deal with the old progressbar on the main branch. There is a better implementation (together with other changes) sitting on the reflink branch waiting to be merged. It replaces the progressbar with https://github.com/cheggaaa/pb and uses an interface to wrap it, much like you did on master.

2) Could you add a short comment on `ChunkStorage` to help with godoc?

3) I really like the idea of separating chunk storage out into its own component, but aren't yet convinced that this needs to deal with the index itself. What I mean is that `ChunkStream()` can easily build it's own chunk array in the right order in its loop and doesn't need `ChunkStorage` to build it. And `ChopFile()` already has the index as input so again, no need for `ChunkStorage` to do anything with it. Without having thought very long about it, I believe the code would become a whole lot simpler if:
    - `ChunkStorage` doesn't know anything about the index but instead focuses on storing chunks that are fed into it (ChunkID + []byte) via channel. Then there's no need to have []byte in `IndexChunk` and it removes the need for the chunk counter
    - `ChunkStream()` builds it's own chunk array to be used in the index
    - `ChunkStorage.getResults()` is re-purposed to only deal with the error and renamed to `s.Error()`

4) There's one optimization in `ChopFile()` I was planning to implement. Right now it iterates over the whole index, checking if each of the chunks is present already before storing. That is quite inefficient since there could be a considerable number of duplicate chunks in an index and there's no need to check over and over if that's already there. The optimization would be to simply dedup the index in a map (ChopFile doesn't care about order) and then store the chunks only once. 3) above would make that quite easy without confusing the index that's kept in `ChunkStorage`.

angelcar:
Alright, 

I think I addressed all your comments, chunk storage is unaware of index now. I think I got away with deleting the call to `ChopFile` from `make`.  I retested everything and it looks fine, but this change  definitely touches a lot of code so take your time to review it and hopefully we can merge it at some point.

P.S. Not sure if this messes up your reflinks branch.

chunkstorage.go
```diff
@@ -3,71 +3,47 @@ package desync
...
+// until there's an error, in which case it will return it.
+func NewChunkStorage(ctx context.Context, cancel context.CancelFunc, n int, ws WriteStore, in <-chan Chunk, pb ProgressBar) *ChunkStorage {
+	s := &ChunkStorage{
+		ctx:    ctx,
+		cancel: cancel,
+		n:      n,
+		ws:     ws,
+		in:     in,
+		pb:     pb,
+		stored: make(map[ChunkID]bool),
```
folbricht: I believe it's a tiny bit more efficient to use a map[ChunkID]struct{} here. Not a big deal though

chunkstorage.go
```diff
@@ -84,111 +60,98 @@ func (s *ChunkStorage) Start() {
...
+	s.Lock()
+	defer s.Unlock()
+	s.stored[c.ID] = true
+}
+
+// Stores a single chunk in a synchronous manner.
+func (s *ChunkStorage) StoreChunk(c Chunk) error {
+
+	// Check in-memory cache to see if chunk has been stored, if so, skip it
+	if s.stored[c.ID] {
```
folbricht: Probably meant to be isChunkStored() to avoid a race

make.go
```diff
@@ -152,33 +165,34 @@ func (c *pChunker) start(ctx context.Context) {
...
 
 		// Check if the next worker already has this chunk, at which point we stop
 		// here and let the next continue
-		if c.next != nil && c.next.syncWith(chunk) {
+		if c.next != nil && c.next.syncWith(ci) {
 			return
 		}
 
+		if c.s != nil {
+			c.err = c.s.StoreChunk(chunk)
```
folbricht: The chunks produced during parallel chunking aren't confirmed yet and shouldn't be stored yet.

chunker.go
```diff
@@ -209,7 +210,8 @@ func (c *Chunker) split(i int, err error) (uint64, []byte, error) {
 	// reset the hash
 	c.hIdx = 0
 	c.hValue = 0
-	return start, b, err
+	id := sha512.Sum512_256(b)
```
folbricht: Calculating the checksum here, rather than in the worker processes is going to make the chunking process much slower since this is done in the main goroutine.

chunkstorage.go
```diff
@@ -3,71 +3,47 @@ package desync
...
 	"github.com/pkg/errors"
 )
 
-type ChunkJob struct {
-	num   int
-	chunk IndexChunk
-}
-
 type ChunkStorage struct {
 	sync.Mutex
```
folbricht: sync.RWMutex would likely a little better here to avoid blocking during concurrent reads

folbricht:
The reflink branch is about de-duping chunks on disk by using seed files and it focuses on the `extract` command that assembles files from indexes. Apart from a few small conflicts with the progressbar that's not going to be an issue here.

Added a few comments in the review above. There are two in particular which I'm not sure can actually be addressed without negating the benefits of the attempt to dedupe this chunk storage code.

- The checksum calculation used to be done in concurrent workers in ChunkStream but now moved to https://github.com/angelcar/desync/blob/f47fbba25594ff5607cbb4cdb4c2a105c45bc0be/chunker.go#L213 which is going to be a performance problem as it's now in the "hot path" of the chunker. Previously the chunker just produced the starting position and a byte slice while checksum calculation/compression/storage was done concurrently.

- The parallel chunking algorithm works by breaking up the file into equal parts (default n=10) and then start chunking recording only the split points along the way. Once they align with the previous worker, those are confirmed and can become chunks. Until that happens, those chunks are not yet valid and can't be stored. This is arguably the most complex part of the code base. Because of how the algorithm works, I'm not sure that the `make` command can be combined into a single operation, it'll need to stay two distinct operations. One that produces the index, then another one to chop up the file based on the index. The effect of https://github.com/angelcar/desync/blob/f47fbba25594ff5607cbb4cdb4c2a105c45bc0be/make.go#L192-L194 will be that additional chunks will be stored, depending on the input data considerably more chunks. When you run the make command now, you'll see the Chunks produced and Overhead at the end. Overhead is the number of chunks that were produced by the algorithm but discarded since they didn't line up. That number can be very large, for example for a file full of 0 bytes where no real split points are produced.

Thinking about this a bit more, we may need to revisit the assumption that this is duplicated code which can be consolidated. Let's look at the input and output of each operation:

### ChunkStream()
- Input:
  - split points + byte slice of chunk data
- Processing:
  - Calculate checksum (ID)
  - Compress chunk data
  - Store chunk
- Output:
  - index

### Chop()
- Input:
  - filename + index
- Processing
  - Read chunk data from file
  - Calculate checksum (ID)
 -  Compare ID with expected from index
  - Compress chunk data
  - Store chunk
- Output:
  - none

While the code looks like a duplicated, it's actually handling different input and producing different output. There's some overlap in the processing stages for sure though. Consolidating this seems harder the longer I think about it. Thoughts?

chunkstorage.go
angelcar: Zero bytes!, I'll take it :)

make.go
angelcar: I'll revert this in the next rev.

chunker.go
angelcar: Oh boy. Silly me, I got overly excited with the encapsulation and totally overlooked this. Reverting this in next rev.

angelcar:
Oh well, I think you're right and this is not entirely duplicated code. I am just encapsulating the Compression + Decompression check + Actual storage. I'm also leaving chunk id dedupping code using the RWMutex. This has a minimal impact in processing performance and, of course, big gain when there are duplicated chunks.

Regarding the `make` command, thanks for clarifying how it works, I reverted that code back. It was too good to be true to make it work in a single pass as opposed to IndexFile + ChopFile. Maybe I'll revisit that at some point but at least I see that it's not easy :(

Let me know what you think of this last rev.

folbricht:
Thank you very much for all the work that's gone in this. Sorry that some of it had to be reverted. Looking cleaner now for sure.

PR #70, opened 2018-11-01
update idFromPath to be OS path agnostic

codingonHP:
On windows, idFromPath function would never return chunk id because of the mismatch of path separator from URL ("/") and Windows file system separator ("\\").
- Use FromSlash that returns the result of replacing each slash ('/') character in the path with a separator character to update the path passed so that path is always in the format of the OS
- Update the function to use os.PathSeparator from os package that takes care of file path separator instead of "/" hardcoding.

### DIFF
--- a/httphandler.go
+++ b/httphandler.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"path/filepath"
 	"strings"
 
@@ -107,6 +108,7 @@ func (h HTTPHandler) put(id ChunkID, w http.ResponseWriter, r *http.Request) {
 }
 
 func (h HTTPHandler) idFromPath(path string) (ChunkID, error) {
+	path = filepath.FromSlash(path)
 	ext := CompressedChunkExt
 	if h.Uncompressed {
 		if strings.HasSuffix(path, CompressedChunkExt) {
@@ -118,9 +120,11 @@ func (h HTTPHandler) idFromPath(path string) (ChunkID, error) {
 	if len(sID) < 4 {
 		return ChunkID{}, fmt.Errorf("expected format '/<prefix>/<chunkid>%s", ext)
 	}
-
+        
+	ps := string(os.PathSeparator)
+	expectedPath := filepath.Join(ps, sID[0:4], sID+ext)
 	// Make sure the prefix does match the first characters of the ID.
-	if path != filepath.Join("/", sID[0:4], sID+ext) {
+	if path != expectedPath {
 		return ChunkID{}, fmt.Errorf("expected format '/<prefix>/<chunkid>%s", ext)
 	}
 	return ChunkIDFromString(sID)


codingonHP:
Sure, I was just getting familiar with the dsync api and so I had set up both store and index on the local file system (windows). I noticed that the chunk server was always throwing route error "expected format '/<prefix>/<chunkid>". On debugging I found out that the path value is having \ as a separator (because chunk was a local folder on C:\)
command
`desync.exe extract -s http://localhost:9000/store/ .\index\xEnviroLog.caibx file.zip`
Error:
`Error: http://localhost:9000/store/: unexpected status code 400 from 2375\23750213743987a6dfe0d4178cd163d3df6f6595d8d2bf82782fe6db711975ec.cacnk`

Same error comes up if I run both chunk-server and index-server and try to get the chunks to client.
I hope the explanation helps

folbricht:
Thank you. That helped. I believe the real bug in this case was on the client. It was using `filepath` instead of `path` which meant it was sending backslash separators over HTTP which is wrong. The server was also using `filepath` to compare against expected path, again that's wrong since all HTTP should use is slash. I made a new PR https://github.com/folbricht/desync/pull/71 for the fix. Would you be able to try that out and see if that addresses your use case?

Btw, in your 2nd command (extract), you do not need to specify the path in the URL, http://localhost:9000/ is sufficient since the chunk-server command will host the chunks under /<index>/<id>.cacnk without the /store prefix

codingonHP:
hey I just tested PR #71 but it did not work as expected and it is still failing with the same issue
commands
```
./desync.exe chunk-server -s ./store/ -l :9000
./desync.exe index-server -s ./index/ -l :9001
 ./desync.exe extract -s http://localhost:9000 http://localhost:9001/index/enviro.caibx file.zip
```

**output**
[==============================================================================================================================================================] 100.00% 0s
Error: http://localhost:9000: unexpected status code 400 from 2375\23750213743987a6dfe0d4178cd163d3df6f6595d8d2bf82782fe6db711975ec.cacnk

I logged the parameters
**path is:** /2375\23750213743987a6dfe0d4178cd163d3df6f6595d8d2bf82782fe6db711975ec.cacnk
**path.Join("/", sID[0:4], sID+ext) is** /2375/2375\23750213743987a6dfe0d4178cd163d3df6f6595d8d2bf82782fe6db711975ec.cacnk

If you want I can look into it

folbricht:
I can't reproduce the problem using the same commands. Can you double-check you're on the `pr-70` branch since that PR hasn't been merged yet. There should be no backslashes in the chunk request, be sure to use an updated client as well as the server. If you're still having this issue, confirm the code you're building looks like https://github.com/folbricht/desync/blob/pr-70/remotehttp.go#L238. Once it's all working, I'll merge it to master.

In your example `http://localhost:9001/index/enviro.caibx` can really be shortened to `http://localhost:9001/enviro.caibx`. The extra path segment will be ignored.

codingonHP:
I can confirm that I am on correct branch because I can see the code changes. I will try again and let you know

codingonHP:
I just tested the code again and it works as expected. The reason it didn't work the last time was because of the fact that I was running the commands in `bash for windows`. Once I switched to `cmd`, paths are resolving correctly. Thoughts?

folbricht:
I don't have bash for windows here to try out unfortunately. It works fine in the Git bash on Windows if I use `/` in the command line. The PR's been merged to master since it addresses a real problem you raised above.

Are you able to find out where the `\` comes from in your env? There should be none of those in HTTP paths.

codingonHP:
Yes I agree that PR addressed the real issue. Thanks :) I will look into the other issue and raise another issue (if it really is) later

folbricht:
Handled in https://github.com/folbricht/desync/pull/71. Closing this for now. Please file a new issue if you find out why it's failing in bash for windows. Thank you

PR #79, opened 2018-11-26
Implement progressbar for untar

kmeaw:
Running "untar" operation from a remote chunk store could be time consuming.
It is tempting for the user to abort the running operation.
This patchset adds a progress bar driven by the number of chunks.

### DIFF
--- a/cmd/desync/progressbar.go
+++ b/cmd/desync/progressbar.go
@@ -20,7 +20,7 @@ func NewProgressBar(prefix string) desync.ProgressBar {
 	return ProgressBar{bar}
 }
 
-// ProgressBar wraps https://github.com/cheggaaa/pb and implrments desync.ProgressBar
+// ProgressBar wraps https://github.com/cheggaaa/pb and implements desync.ProgressBar
 type ProgressBar struct {
 	*pb.ProgressBar
 }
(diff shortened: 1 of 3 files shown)


cmd/desync/progressbar.go
```diff
@@ -39,3 +39,48 @@ func (p ProgressBar) Start() {
...
+		return nil
+	}
+
+	cur, err := f.Seek(0, os.SEEK_CUR)
+	if err != nil {
+		return nil
+	}
+
+	size, err := f.Seek(0, os.SEEK_END)
+	f.Seek(cur, os.SEEK_SET)
```
folbricht: Missing error handling?

cmd/desync/progressbar.go
```diff
@@ -39,3 +39,48 @@ func (p ProgressBar) Start() {
...
+	return &pr
+}
+
+func (p ProgressReader) Read(b []byte) (n int, err error) {
+	n, err = p.f.Read(b)
+	if err != nil {
+		return
+	}
+
+	cur, err := p.f.Seek(0, os.SEEK_CUR)
```
folbricht: This could be handled by incrementing the progress bar by `n` rather than the seek

cmd/desync/untar.go
```diff
@@ -66,7 +66,10 @@ func runUntar(ctx context.Context, opt untarOptions, args []string) error {
 			return err
 		}
 		defer f.Close()
-		return desync.UnTar(ctx, f, targetDir, opt.UntarOptions)
+		pb := NewProgressReader("Unpacking ", f)
+		pb.Start()
```
folbricht: pb can be nil, so would need to wrap that in `if pb != nil` to avoid panic when STDERR is not a terminal (like when redirecting output with 2>file)

untar.go
```diff
@@ -211,8 +211,12 @@ func UnTarIndex(ctx context.Context, dst string, index Index, s Store, n int, op
 	// Feeder - requesting chunks from the workers and handing a result data channel
 	// to the assembler
 	g.Go(func() error {
+                pb.SetTotal(len(index.Chunks))
```
folbricht: Also needs an `if` to avoid nil ptr

folbricht:
Thanks a lot for the contribution. Like the way you added the progress bar based on bytes read. In this case the use of `Seek()` to determine the file size seems a bit over the top though. Also the progress bar can be made to record bytes written (or read) without introducing another type.

I tried to simplify some of this in https://github.com/folbricht/desync/pull/82. Would you be able to take a look if that meets the requirements or if I missed something there?

kmeaw:
Thank you for the review.
I have taken a look at #82 and feel your implementation is better than mine.

folbricht:
Handled in https://github.com/folbricht/desync/pull/82. Thanks again for this.

PR #80, opened 2018-11-26
tar: add one-file-system flag

kmeaw:
GNU tar has a --one-file-system flag which is useful for tarring filesystems.
Desync's tar recursively traverses into the mountpoints which could be not desirable.
This patchset adds an option to exclude all mountpoints from the index.

### DIFF
--- a/tar.go
+++ b/tar.go
@@ -28,17 +28,24 @@ const TarFeatureFlags uint64 = CaFormatWith32BitUIDs |
 
 // Tar implements the tar command which recursively parses a directory tree,
 // and produces a stream of encoded casync format elements (catar file).
-func Tar(ctx context.Context, w io.Writer, src string) error {
+func Tar(ctx context.Context, w io.Writer, src string, oneFileSystem bool) error {
 	enc := NewFormatEncoder(w)
 	info, err := os.Lstat(src)
 	if err != nil {
 		return err
 	}
-	_, err = tar(ctx, enc, src, info)
+	dev := uint64(0)
+	if oneFileSystem {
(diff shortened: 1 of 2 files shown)


tar.go
```diff
@@ -96,6 +110,14 @@ func tar(ctx context.Context, enc FormatEncoder, path string, info os.FileInfo)
 		}
 		var items []FormatGoodbyeItem
 		for _, s := range stats {
+			if dev != 0 {
+				// one-file-system is set, skip other filesystems
+				st, ok := s.Sys().(*syscall.Stat_t)
+				if ok && st.Dev != dev {
```
folbricht: If we get here and Sys() is not *syscall.Stat_t, shouldn't we `continue` as well? Since it sort of indicates a different filesystem of sorts which is different from the one in `dev`.

tar.go
```diff
@@ -54,11 +62,17 @@ func tar(ctx context.Context, enc FormatEncoder, path string, info os.FileInfo)
 	)
 
 	switch sys := info.Sys().(type) {
+	case *arctar.Header:
```
folbricht: Not quite sure where this comes from. Could you explain a bit more of when that code path would be used?

tar.go
kmeaw: You are right, it should be `!ok || st.Dev != dev` instead of `ok && st.Dev != dev`.

tar.go
kmeaw: It was an initial attempt of another feature's implementation: allow desync to use a .tar file instead of filesystem tree as a "tar" source (issue #84).

PR #81, opened 2018-11-26
S3: allow user to specify lookup type

kmeaw:
minio supports two ways to address S3 buckets: DNS- and path-style lookups (http://bucket.s3endpoint.org vs http://s3endpoint.org/bucket).
Desync does not have a way to specify the type of bucket addressing, like boto3's addressing_style configuration key, and minio falls back to the autodetection method (which actually works only for Google and Amazon, not for private installations).
This patchset gives the user a way to specify the method by appending "?lookup=dns", "?lookup=path" or "?lookup=auto" to the S3 path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,18 @@ This is a store running on the local machine on port 9000 without SSL.
 s3+http://127.0.0.1:9000/store
 ```
 
+#### Setting S3 bucket addressing style for other services
+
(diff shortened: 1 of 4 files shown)


folbricht:
Thanks for the PR. I'm not able to test this myself but it looks pretty good. Would you be able to add something to the documentation under the S3 documentation in README.md (https://github.com/folbricht/desync/blob/master/README.md#s3-chunk-stores)? Maybe add an example to the list at the end of the file. It really helps.

PR #85, opened 2018-12-06
Resolve #83: implement extended attributes (xattrs)

### LINKED ISSUE #83: enhancement : add xattrs support
Currently desync does not support extended attributes.
Extended attributes support is needed to use desync as a root filesystem imaging tool.

kmeaw:
Most POSIX filesystems allow extended attributes to be assigned to inodes.
Currently desync does not support extended attributes.
This patch implements extended attributes in an casync-compatible way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Among the distinguishing factors:
 - Supports local stores as well as remote stores (as client) over SSH, SFTP and HTTP
 - Built-in HTTP(S) chunk server that can proxy multiple local or remote stores and also supports caching.
 - Drop-in replacement for casync on SSH servers when serving chunks read-only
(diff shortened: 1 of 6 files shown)


charles-dyfis-net:
Is this new dependency available on all previously-supported platforms (including MacOS and Windows)?

kmeaw:
New dependency is available on darwin,bsd,linux. It is not available on windows, however this platform is not supported by tar.go and untar.go.

folbricht:
Thanks for the PR. Just a couple of comments:

- Any error from `xattr.Lset()` is currently ignored. Is that deliberate? 

- README.md says xattrs aren't supported. Since you implemented that, could you remove that reference?

kmeaw:
Now errors from LSet are handled. However I put the LSet code under `!opts.NoSameOwner` condition — to make sure `desync untar` could be run under a non-privileged user.

Issue #88, opened 2018-12-30
"desync tar" can bail outright on a UID lookup failure


charles-dyfis-net:
```
# desync tar out.catar /mnt
user: lookup userid 0: no such file or directory
```

The above was observed in a sandboxed environment without an `/etc/passwd`.

If `/etc/passwd` is created but empty, the result is a bit different:

```
user: unknown userid 0
```

I would expect desync to continue (with numeric IDs only) should username lookup fail.

folbricht:
The failure isn't actually related to `tar` at all. It happens when desync tries to find the default config file location which involves a lookup of the current user. PR https://github.com/folbricht/desync/pull/89 should address this, but I'm not entirely sure since I was only able to repro one of the failures. A retry with that fix would be much appreciated.

charles-dyfis-net:
Huh. That's... interesting.

As a user, when I change the value of `HOME`, I expect tools to honor that value from the environment and not try to perform an account-database lookup... but then again, maybe that doesn't make sense on Windows?

charles-dyfis-net:
(Which is to say that my proposed fix would be to use `${HOME:-}/.config/desync`, and skip the account-database look entirely).

folbricht:
That does seem like the simpler solution, though there'll need to be some OS-specific code since the equivalent on Windows is `HOMEDRIVE` + `HOMEDIR`. Going to update the PR accordingly.

folbricht:
Resolved in https://github.com/folbricht/desync/pull/89

PR #89, opened 2019-01-01
Don't fail on user lookup when determining the default config file lo…

folbricht:
…cation as per #88


charles-dyfis-net:
I'd probably address this differently, but the proposed change *does* resolve the issue in both empty and nonexistent /etc/passwd cases.

folbricht:
This implementation now uses HOME (HOMEDIR+HOMEPATH on Windows) and no longer reads the user database

PR #94, opened 2019-01-23
Fix untar permissions and error handling

kmeaw:


untar.go
```diff
@@ -208,6 +208,7 @@ func UnTarIndex(ctx context.Context, dst string, index Index, s Store, n int, op
 		req      = make(chan requestJob)
 		assemble = make(chan chan []byte, n)
 	)
+	ctx, cancel := context.WithCancel(ctx)
```
folbricht: Going to need a `defer cancel()` here to avoid a context leak.

untar.go
```diff
@@ -284,6 +294,11 @@ func UnTarIndex(ctx context.Context, dst string, index Index, s Store, n int, op
 	// Run untar in the main goroutine
 	err := UnTar(ctx, r, dst, opts)
 
+	if err != nil {
```
folbricht: With the `defer cancel()` in place above, all of this (to the end of the function) could be written as:
```go
	// Run untar in the main goroutine
	if err := UnTar(ctx, r, dst, opts); err != nil {
		return err
	}
	return g.Wait()
```

But thinking a bit more about this, what do you think about running the UnTar() inside a goroutine as well with g.Go() like the others? Then we wouldn't need the extra context, and there's no question about which error gets priority. All would be covered by `return g.Wait()` at the end. That should make things a little simpler.

kmeaw: Thank you for such a nice idea; done.

folbricht:
Thanks for those fixes.

PR #95, opened 2019-01-29
Fixes issue with 32 bit builds of desync

svishwanath-tw:
Alignment issues with 64 bit fields in the struct causes 32 bit builds of desync to not work (panic in sync/atomic package).
Re-arranging the members of the struct fixes this problem
Based on suggestions found at https://github.com/golang/go/issues/599 and https://github.com/golang/go/issues/23345.

### DIFF
--- a/extractstats.go
+++ b/extractstats.go
@@ -7,15 +7,15 @@ import (
 // ExtractStats contains detailed statistics about a file extract operation, such
 // as if data chunks were copied from seeds or cloned.
 type ExtractStats struct {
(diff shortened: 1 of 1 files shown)


svishwanath-tw:
Discovered the problem and solution when trying to run this tool on a 32 bit machine.

folbricht:
Thank you. I didn't know that, even though it actually says as much in https://golang.org/pkg/sync/atomic/#pkg-note-BUG

PR #115, opened 2019-08-10
Use a BST for the list of Goodye items

### LINKED ISSUE #113: Directory serialization order.
The FormatGoodbyeItem's in tar.go places things in a different order than casync. This may cause reproducibility issues as well as errors with mount. 

Looks like there is an issue with the order of files from ioutil.ReadDir, io.File.Readdir and whatever casync uses/expects.

folbricht:
As per #113, desync until now did not use the same order of the Goodbye items in catar archives preventing casync from FUSE-mounting such archives. This PR implements the BST algorithm casync uses.

Closes #113

### DIFF
--- a/format.go
+++ b/format.go
@@ -8,6 +8,7 @@ import (
 	"math"
 	"os"
 	"reflect"
+	"sort"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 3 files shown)


tomberek:
It works. Tested with largish directory.

Issue #114, opened 2019-08-09
CaFormatExcludeFile


tomberek:
With #113 resolved, the remaining differences between casync and desync produced .caidx/.catar files are flags. Most can be resolved by being very explicit with casync:

```
casync make --with=unix \
                      --with=xattrs \
                      --with=nsec-time \
                      --without=16bit-uids \
                      --with=32bit-uids \
                      --chunk-size=16384:65536:262144 \
                      casync.catar /data-ro/temp
```
The last flag missing is CaFormatExcludeFile. With these changes and a correct file order, produced caidx/catar are bit-identical to casync.

PR #116, opened 2019-08-11
Add CaFormatExcludeFile flag to catar archives

folbricht:
This adds CaFormatExcludeFile to feature flags as per #114. The catar flags produced for both are now: `0xb000000010001f22`.

### DIFF
--- a/const.go
+++ b/const.go
@@ -71,6 +71,7 @@ const (
 	CaFormatWithSELinux = 0x40000000
 	CaFormatWithFcaps   = 0x80000000
 
+	CaFormatExcludeFile      = 0x1000000000000000
 	CaFormatSHA512256        = 0x2000000000000000
 	CaFormatExcludeSubmounts = 0x4000000000000000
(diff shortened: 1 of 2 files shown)


tomberek:
Not sure if this should be added if the feature doesn't actually exist in desync. Or does it?

folbricht:
I'm not even sure what that flag exactly means and why it's added to archives by casync.

tomberek:
It it similar to .gitignore. Allows you to exclude files from a directory tree. See documentation at the end of https://github.com/systemd/casync/blob/master/doc/casync.rst

folbricht:
That flag seems to refer more to a capability of the tool that produced the archive, not something the archive itself needs to know. casync adds that even without actually excluding files. So no, desync doesn't currently support this, but I'm ok adding it if that aligns better with casync.

tomberek:
Let's leave the issue open to add the capability in the future, but merge this to get closer for reproducibility/testing?

PR #121, opened 2019-08-18
Support SHA256 digest algorithm

### LINKED ISSUE #120: SHA256 support
- casync supports it
- other major tech in this space does not support SHA512/256, like IPFS, webtorrent, multihash. It would be nice to just directly use the hashes in indexes to generate IPFS addresses or magnet links in order to exchange chunks.

folbricht:
Implements additional SHA256 hashing algorithm. It's set globally with --digest=sha256

Closes #120


tomberek:
Did some basic testing/parsing. Seems to work well.

PR #130, opened 2019-09-29
Switch to using v2 Fuse library

folbricht:
Switch to using the newer v2 API of the FUSE library (github.com/hanwen/go-fuse/v2/fs). Not only is that simpler to use, but also let's the kernel do more caching.


limbo127:
ok, i'm trying to test as soon as possible
thanks
Nicolas

PR #131, opened 2019-10-15
Add ability to read tar archives that lack common root directory

astsmtl:
Add command-line option '--tar-add-root'. If enabled desync will pretend that
all top-level tar archive elements are in one top-level root directory. Helps
for ex. while running 'desync tar' on result of 'docker export'.

### DIFF
--- a/cmd/desync/tar.go
+++ b/cmd/desync/tar.go
@@ -20,6 +20,7 @@ type tarOptions struct {
 	createIndex bool
 	desync.LocalFSOptions
 	inFormat string
+	desync.TarReaderOptions
 }
 
 func newTarCommand(ctx context.Context) *cobra.Command {
@@ -50,6 +51,7 @@ the input can be a tar file or stream to STDIN with '-'.
 	flags.BoolVarP(&opt.OneFileSystem, "one-file-system", "x", false, "don't cross filesystem boundaries")
 	flags.StringVar(&opt.inFormat, "input-format", "disk", "input format, 'disk' or 'tar'")
 	flags.BoolVarP(&opt.NoTime, "no-time", "", false, "set file timestamps to zero in the archive")
+	flags.BoolVarP(&opt.AddRoot, "tar-add-root", "", false, "pretend that all tar elements have a common root directory")
 	addStoreOptions(&opt.cmdStoreOptions, flags)
 	return cmd
 }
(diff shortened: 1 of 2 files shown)


cmd/desync/tar.go
```diff
@@ -50,6 +51,7 @@ the input can be a tar file or stream to STDIN with '-'.
 	flags.BoolVarP(&opt.OneFileSystem, "one-file-system", "x", false, "don't cross filesystem boundaries")
 	flags.StringVar(&opt.inFormat, "input-format", "disk", "input format, 'disk' or 'tar'")
 	flags.BoolVarP(&opt.NoTime, "no-time", "", false, "set file timestamps to zero in the archive")
+	flags.BoolVarP(&opt.tarAddRoot, "tar-add-root", "", false, "pretend that all tar elements have common root directory")
```
folbricht: "...have **a** common..." ?

tarfs.go
```diff
@@ -100,39 +101,54 @@ func (fs TarWriter) Close() error {
...
 }
 
-var _ FilesystemReader = TarReader{}
+var _ FilesystemReader = &TarReader{}
 
 // NewTarFS initializes a new instance of a GNU tar archive that can be used
 // for catar archive tar/untar operations.
-func NewTarReader(r io.Reader) TarReader {
-	return TarReader{gnutar.NewReader(r)}
+func NewTarReader(r io.Reader, addRoot bool) *TarReader {
```
folbricht: Could you add a new struct `TarReaderOptions` and put the bool into that one? That would make it more flexible if more additional fields have to be added later. Take a look at https://github.com/folbricht/desync/blob/master/localfs.go#L53 and https://github.com/folbricht/desync/blob/master/cmd/desync/untar.go#L18

tarfs.go
```diff
@@ -100,39 +101,54 @@ func (fs TarWriter) Close() error {
...
-		Gid:        h.Gid,
-		Xattrs:     h.Xattrs,
-		DevMajor:   uint64(h.Devmajor),
-		DevMinor:   uint64(h.Devminor),
-		Data:       ioutil.NopCloser(fs.r),
+func (fs *TarReader) Next() (f *File, err error) {
+	if fs.root != nil {
+		f = fs.root
+		fs.root = nil
+	} else {
```
folbricht: Might be better to add a return above and remove the `else` to reduce nesting of the block below

PR #135, opened 2019-10-30
Basic untar support for Windows platform

dan-embark:
Windows not supports the untar command.

Limitations:
- Device entries are not supported and will terminate the untar operator with an error
- --no-same-owner and --no-same-permissions are not supported, permission settings are ignored


folbricht:
Thanks for the PR. Rather than duplicating the untar logic, wouldn't it be better to separate the platform specific pieces like https://github.com/folbricht/desync/blob/master/localfs.go#L167 out into separate files with one for Windows that's a basicall a NOP or prints a warning?

folbricht:
Thanks, I should be able to help out later today if you get stuck. Might create a branch for us to work on.

dan-embark:
Also, I see now there are some mistakes in there, will clean it up

localfs_windows.go
```diff
@@ -0,0 +1,125 @@
+// +build windows
```
folbricht: - Since the filename is already suffixed with `_windows` the extra build tag isn't actually needed here.

- I believe localfs.go needs build tags to exclude windows, or the build will fail with name conflicts on Windows.

localfs_windows.go
```diff
@@ -0,0 +1,125 @@
...
+		return err
+	}
+	if err := os.Symlink(n.Target, dst); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (fs *LocalFS) CreateDevice(n NodeDevice) error {
+	return errors.New("Not available on this platform")
```
folbricht: Could you clarify that it's the device that's not supported. Like `device nodes not supported on this platform`.

localfs_windows.go
```diff
@@ -0,0 +1,125 @@
...
+
+type walkEntry struct {
+	path string
+	info os.FileInfo
+	err  error
+}
+
+// Next returns the next filesystem entry or io.EOF when done. The caller is responsible
+// for closing the returned File object.
+func (fs *LocalFS) Next() (*File, error) {
```
folbricht: There's nothing stopping us from implementing this as well on Windows. I can add that in later.

localfs_windows.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// Ignore the incoming permissions when writing files. Use the current default instead.
+	NoSamePermissions bool
+
+	// Reads all timestamps as zero. Used in tar operations to avoid unneccessary changes.
+	NoTime bool
+}
+
+var _ FilesystemWriter = &LocalFS{}
+var _ FilesystemReader = &LocalFS{}
```
folbricht: I'm wondering if we could move the code that's common between OS' into a file that's included in all builds to avoid the duplication. Basically have three files:
- localfs.go - common to all builds, would contain things like the structures
- localfs_windows.go - Windows specific file creation and reading
- localfs_other.go - Non-windows via build tags in the header

localfs_windows.go
dan-embark: Yes, I did not spend time on it since we do not have support for tar on Windows yet.

dan-embark:
A bit cleaner yet, broke out Set*Permissions parts of CreateDir, CreateFile and CreateSymLink since those contain the platform specific information.
CreateDisk is still entirely in the platform specific files since they don't have as a specific part to break out.
Intentionally did not spend time of fs Next() since it is not needed for untar support and trying to keep the scope down.

Thanks for the feedback @folbricht

folbricht:
Thank you

Issue #144, opened 2020-01-29
Retrieve AWS credentials from ~/.aws/credentials


Joe-Improbable:
Currently desync doesn't fall back to using `~/.aws/credentials` if no other methods of providing credentials is used.

The code for falling back to the default credential location exists (`credentials.go:103` `Retrieve()`) however it is not used. 

Is there reason this functionality isn't used?

folbricht:
Resolved in https://github.com/folbricht/desync/pull/145

PR #145, opened 2020-01-29
Fallback to AWS default creds if none are found

Joe-Improbable:
Solves #144

### DIFF
--- a/cmd/desync/config.go
+++ b/cmd/desync/config.go
@@ -62,16 +62,8 @@ func (c Config) GetS3CredentialsFor(u *url.URL) (*credentials.Credentials, strin
 	// if access access-key is present, it takes precedence
 	if credsConfig.AccessKey != "" {
 		creds = NewStaticCredentials(credsConfig.AccessKey, credsConfig.SecretKey)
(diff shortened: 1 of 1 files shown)


cmd/desync/config.go
```diff
@@ -64,6 +64,14 @@ func (c Config) GetS3CredentialsFor(u *url.URL) (*credentials.Credentials, strin
 		creds = NewStaticCredentials(credsConfig.AccessKey, credsConfig.SecretKey)
 	} else if credsConfig.AwsCredentialsFile != "" {
 		creds = NewRefreshableSharedCredentials(credsConfig.AwsCredentialsFile, credsConfig.AwsProfile, time.Now)
+	} else {
```
folbricht: I believe you could just change the above condition from
```go
} else if credsConfig.AwsCredentialsFile != "" {
```

to:
```go
} else {

That that would have the same effect, without the extra `else`, wouldn't it?
```

Joe-Improbable: Good spot, I just pushed that change

folbricht:
Thanks for the fix!

PR #147, opened 2020-02-05
Use the correct home directory on windows #146

### LINKED ISSUE #146: Bug - Windows does not find AWS credentials
The user home directory is got using `home := os.Getenv("HOME")` which many return the empty string on some versions of windows.

This means on Windows s3 credentials will not always be loaded from the config file.

I've raised a PR to fix this

Let me know what you think

Joe-Improbable:
Fixes #146

### DIFF
--- a/cmd/desync/credentials.go
+++ b/cmd/desync/credentials.go
@@ -17,14 +17,20 @@ import (
 // Builds the shared config file path based on the OS's platform.
 //
 //   - Linux/Unix: $HOME/.aws/credentials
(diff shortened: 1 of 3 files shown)


folbricht:
That is something that I had on my TODO list but not got around to. Since Go 1.12, we can use https://golang.org/pkg/os/#UserHomeDir I believe. How about using the stdlib function and requiring Go 1.12 in go.mod?

Joe-Improbable:
Good idea, I've pushed these changes.

folbricht:
Thank you

PR #150, opened 2020-02-07
use seperate queues for hasChunk and get/storeChunk

### LINKED ISSUE #149: Windows - Writededupqueue race condition
I believe there is a race condition on Windows when using a Cache.

The root cause is that the `Writededupqueue.go` uses the `getChunkQueue` for both `StoreChunk`, `GetChunk` and `HasChunk`

I believe `HasChunk` should instead be using the currently unused `hasChunkQueue`

The race condition:

| Thread 1        | Thread 2           |
| ------------- |:-------------:|
| HasChunk    |  | $1600 |
| Stores request in requests map (Locks getChunkQueue mutex)  | StoreChunk
| | Request is in flight (Mutex is locked so waits) 
| Finds chunk is missing and returns `chunkMissing` on channel |      |
| removes request from request map  (Unlocks mutex)  |
| | Receives a `chunkMissing` error on the channel and returns causing `cache.GetChunk` to return a `chunkMissing` error

I'll raise a PR for my fix.

Let me know if you need a better explanation

Joe-Improbable:
This should fix #149

### DIFF
--- a/dedupqueue.go
+++ b/dedupqueue.go
@@ -57,7 +57,7 @@ func (q *DedupQueue) GetChunk(id ChunkID) (*Chunk, error) {
 }
 
 func (q *DedupQueue) HasChunk(id ChunkID) (bool, error) {
-	req, isInFlight := q.getChunkQueue.loadOrStore(id)
+	req, isInFlight := q.hasChunkQueue.loadOrStore(id)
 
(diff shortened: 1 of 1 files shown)


folbricht:
Nice find. I missed that.

PR #151, opened 2020-02-10
Use seperate queues for get/store/has chunk

Joe-Improbable:
This is the follow up to #149 (and #150)

It solves the remaining race condition between `GetChunk` and `StoreChunk`

This happens in case that `GetChunk` runs before `StoreChunk`, this solve that by separating those requests into separate queues.

However, in the case `GetChunk` occurs whilst a `StoreChunk` happens, we can just wait for `StoreChunk` to return us the chunk, rather than launching a new request

### DIFF
--- a/writededupqueue.go
+++ b/writededupqueue.go
@@ -1,5 +1,7 @@
 package desync
 
+import "fmt"
+
 var _ WriteStore = &WriteDedupQueue{}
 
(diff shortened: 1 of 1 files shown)


folbricht:
Thanks for that fix. I'm going to add a test for it as well.

Issue #159, opened 2020-04-22
Retry with delay and backoff for some HTTP responses


Kalmalyzer:
I am using desync against Google Cloud Store. I configure GCS's REST API as a chunk store. If I upload 100MB of test data then the GCS backend responds with a 502 Bad Gateway partway through the upload.

I'm not sure why; I think it is because desync attempts to push more data than GCS can handle via a single TCP connection. Regardless of the reason, would it make sense to have a retry mechanism on some failures? (most/all of 5xx error codes are potentially transient and should be retryable.)

For example, https://github.com/hashicorp/go-retryablehttp implements a default backoff model of N seconds delay before the Nth attempt.

(Perhaps it would be a good idea to swap out raw net/http with go-retryablehttp rather than fleshing out a homemade retry mechanism?)

Kalmalyzer:
One implementation of this is available in #161 .

Kalmalyzer:
There is now an optional retry mechanism in place for HTTP/HTTPS stores.

The code base used to retry network-level failures (DNS lookup failures, TCP connection establishment failures, SSL handshake fail, TCP stream breaking mid-request, ...). #161 extends that retry to apply to 5xx class HTTP responses as well.

Retries are instant by default, but the user can activate a linear backoff (after error N, the corresponding goroutine will sleep for N*base_interval microseconds before retrying).

With the retry mechanism active, I have successfully used Google Cloud Storage's REST API for storing chunks (100MB test suite, tests run a half-dozen times). The intermittent 502s have not resulted in any application-level failures in desync so far.

Closing!

PR #160, opened 2020-04-23
add Google Storage as a backend store

zimbatm:
Would something like that be interesting on principle?

It adds a `gs://` scheme to store both the chunks and indexes on Google Storage.

This is at a very early stage. I just copied the S3 store and adapted it.


Kalmalyzer:
I haven't looked at your actual implementation, just would like to mention that I too am using Google Storage as chunk/index store.

I want authentication to be done to be done using regular Google accounts (the same kind of auth flow that you experience when you do `gcloud auth login`) and not with shared secrets / private keys. For my particular use case, I have written a C# utility that does the OAuth flow, and then generates a config file for desync with the access token as the http-auth token for the Google APIs. Then I point desync to the Google Storage REST API endpoint.

The auth bit itself is a lot of wrangling. It only provides tokens that are valid for 1 hour (so the desync invocation will start to fail its HTTPs requests after an hour) and there are occasional HTTP 502 errors from the backend that require retries (see #159).

I would see value in this if it helped with any of these:
1) it reduced the number of HTTP requests performed
2) it did error handling better than remotehttp.go does today
3) it handled some forms of google-specific credentials in a simple fashion (similar to how s3 credentials are handled in the config file)
4) it made gs:// URLs directly portable between `gsutil` and `desync`

Perhaps it helps with some of those?

zimbatm:
Regarding auth, it uses the blessed google library which sources the auth credentials from the common places such as the metadata server (when running inside of Google Cloud), config files and environment variables. I don't know if it refreshes the auth token in the metadats server case. To be verified.

It also takes the same `gs://my-bucket/prefix` URLs as `gsutil`

zimbatm:
`./desync cache -c gs://bucket-name/prefix -s ./testdata/blob1.store/ ./testdata/blob1.caibx` is working now

folbricht:
I'm not setup test this, but what I see so far is looking pretty good. Let me know when you're comfortable with it and I'll take another look.

Kalmalyzer:
Seconded, this is looking promising. I have done some brief tests tar'ing/untar'ing against a GCS bucket and it was working as intended. (I relied on it using whatever authentication was present on my machine already, as I'm regularly using gcloud.)

zimbatm:
Left two comments which I think should be addressed. We probably want to also allow custom credentials to be passed through the CLI but that could be done in another PR as needed.

folbricht:
Sorry, merged another PR which created a conflict, could you rebase? Should be a simple one

zimbatm:
Rebased.

I don't know that I will spend much more effort on this PR at this point as I don't have a concrete usage for desync yet. From my cursory testing, things are working as expected.

I really like the concept of the tool and might revisit it in the future.

Kalmalyzer:
This looks good, and I hope to begin using this if it gets accepted - so thanks for this :)

folbricht:
Thanks a lot for the PR. Much appreciated. I see what looks like a minor issue in error handling but can fix that after merging easily enough.

@Kalmalyzer Thanks for trying it out. I don't have an env for this at the moment so please let me know if anything comes up.

PR #158, opened 2020-04-22
Verbose logging for remotehttp

Kalmalyzer:
Related to #157 - this provides a "--verbose" commandline switch, and when active, it prints HTTP requests/responses done via the remotehttp module.

The output looks like this:
```
2020/04/22 22:33:06 HTTP HEAD: https://storage.googleapis.com/<bucketname>/8f3b/8f3b7cf12857a73edb96afe45f4dad05bc97695daceec3a54062ad26b5689715.cacnk: request sent
2020/04/22 22:33:06 HTTP HEAD: https://storage.googleapis.com/<bucketname>/8f6d/8f6d1389a95d165e1de44aef9183e0918df3695bc54dde347c08d082b41745f9.cacnk: request sent
2020/04/22 22:33:06 HTTP HEAD: https://storage.googleapis.com/<bucketname>/cc40/cc4065a576a9f79a1d7455e6fdae5dc056befffea3b178a5abd554eada58603d.cacnk: request sent
2020/04/22 22:33:06 HTTP PUT: https://storage.googleapis.com/<bucketname>/2e12/2e127eb6618a59962a8696875cabd9d8f7dde40360724fc6084a90b118583758.cacnk: response received: 200 OK
2020/04/22 22:33:06 HTTP HEAD: https://storage.googleapis.com/<bucketname>/3021/3021d108e13e9a63138f02aa57e2034a0ef1dde46993b086804e0c776154523f.cacnk: request sent
2020/04/22 22:33:06 HTTP HEAD: https://storage.googleapis.com/<bucketname>/5ecc/5ecc236603e745b008a7c66beb939d1e6b2361754878bdc2d2cafc9d3015e679.cacnk: response received: 404 Not Found
2020/04/22 22:33:06 HTTP PUT: https://storage.googleapis.com/<bucketname>/5ecc/5ecc236603e745b008a7c66beb939d1e6b2361754878bdc2d2cafc9d3015e679.cacnk: request sent
2020/04/22 22:33:06 HTTP PUT: https://storage.googleapis.com/<bucketname>/5ecc/5ecc236603e745b008a7c66beb939d1e6b2361754878bdc2d2cafc9d3015e679.cacnk: response received: 502 Bad Gateway
2020/04/22 22:33:06 HTTP PUT: https://storage.googleapis.com/<bucketname>/d440/d440a71089310cf27fda85164f4f21f8122ca2cafdbfff8657677af1ad8b0ac5.cacnk: response received: 200 OK
[...]

PR #161, opened 2020-04-25
Retry 5xx errors

Kalmalyzer:
(This is based on PR #158  - only the two commits on Apr 25th are unique to this PR)

This change provides retry mechanisms for REST API 5xx errors, in addition to the internal errors that were retried. It provides linear backoff during successive retries for a single request. There's a sizable test suite for `remotehttp` as well now. Documentation has been updated to mention the `error-retry-base-interval` setting.

With this change, I have successfully used tar + untar with a 100MB test file set against Google Cloud Storage, and had it work successfully despite one intermittent HTTP 502 response in the middle (this used to make the entire operation fail).

A couple of things are worth pointing out:
- I have centralized all HTTP activity within `remotehttp` to `IssueRetryableHttpRequest` / `IssueHttpRequest`. The logic was getting too complicated to have it duplicated across 3 functions.
- The current implementation will always read the response body, even for HTTP PUT and HTTP HEAD operations. This is wasteful but I wasn't up for doing specialization there. I do not expect that to make a significant performance difference (response bodies should be small for those requests).
- The previous interface expected anyone who called `remotehttp.StoreObject()` to provide an `io.Reader` for the body. I have changed this into a function that returns an `io.Reader`. This way, retrying the same HTTP request will be done with a fresh `io.Reader` (which should emit the same set of bytes as the previous one) for each attempt.
- The `remotehttpindex.StoreIndex()` operation will, when there are intermittent 5xx errors, serialize the index and feed it into a pipe once per attempt. This is wasteful, but the alternative is to serialize all of the index up-front for every request, and keep it in a buffer across all retries. That is wasteful when there are no errors. I have chosen to optimize performance for the no-error case.
- The retry base interval is tweakable per-store, just like the pre-existing max retry count.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,7 @@ Available configuration values:
 - `store-options` - Allows customization of chunk and index stores, for example comression settings, timeouts, retry behavior and keys. Not all options are applicable to every store, some of these like `timeout` are ignored for local stores. Some of these options, such as the client certificates are overwritten with any values set in the command line. Note that the store location used in the command line needs to match the key under `store-options` exactly for these options to be used. Watch out for trailing `/` in URLs.
   - `timeout` - Time limit for chunk read or write operation in nanoseconds. Default: 1 minute. If set to a negative value, timeout is infinite.
   - `error-retry` - Number of times to retry failed chunk requests. Default: 0.
-  - `error-retry-base-interval` - Number of seconds to wait before first retry attempt. Retry attempt number N for the same request will wait N times this interval. Default: 1 second.
+  - `error-retry-base-interval` - Number of seconds to wait before first retry attempt. Retry attempt number N for the same request will wait N times this interval. Default: 0 seconds.
   - `client-cert` - Cerificate file to be used for stores where the server requires mutual SSL.
   - `client-key` - Key file to be used for stores where the server requires mutual SSL.
   - `ca-cert` - Certificate file containing trusted certs or CAs.
--- a/README.md
+++ b/README.md
@@ -238,6 +238,7 @@ Available configuration values:
 - `store-options` - Allows customization of chunk and index stores, for example comression settings, timeouts, retry behavior and keys. Not all options are applicable to every store, some of these like `timeout` are ignored for local stores. Some of these options, such as the client certificates are overwritten with any values set in the command line. Note that the store location used in the command line needs to match the key under `store-options` exactly for these options to be used. Watch out for trailing `/` in URLs.
   - `timeout` - Time limit for chunk read or write operation in nanoseconds. Default: 1 minute. If set to a negative value, timeout is infinite.
   - `error-retry` - Number of times to retry failed chunk requests. Default: 0.
+  - `error-retry-base-interval` - Number of seconds to wait before first retry attempt. Retry attempt number N for the same request will wait N times this interval. Default: 1 second.
   - `client-cert` - Cerificate file to be used for stores where the server requires mutual SSL.
   - `client-key` - Key file to be used for stores where the server requires mutual SSL.
   - `ca-cert` - Certificate file containing trusted certs or CAs.
--- a/remotehttp.go
+++ b/remotehttp.go
@@ -32,6 +32,8 @@ type RemoteHTTP struct {
 	*RemoteHTTPBase
 }
 
+type GetReaderForRequestBody func() io.Reader
+
 // NewRemoteHTTPStoreBase initializes a base object for HTTP index or chunk stores.
 func NewRemoteHTTPStoreBase(location *url.URL, opt StoreOptions) (*RemoteHTTPBase, error) {
 	if location.Scheme != "http" && location.Scheme != "https" {
@@ -96,7 +98,7 @@ func (r *RemoteHTTPBase) String() string {
 func (r *RemoteHTTPBase) Close() error { return nil }
 
 // Send a single HTTP request.
-func (r *RemoteHTTPBase) IssueHttpRequest(method string, u *url.URL, rdr io.Reader, attempt int) (int, []byte, error) {
+func (r *RemoteHTTPBase) IssueHttpRequest(method string, u *url.URL, getReader GetReaderForRequestBody, attempt int) (int, []byte, error) {
 
 	var (
 		resp *http.Response
@@ -107,7 +109,7 @@ func (r *RemoteHTTPBase) IssueHttpRequest(method string, u *url.URL, rdr io.Read
 		})
 	)
 
-	req, err := http.NewRequest(method, u.String(), rdr)
+	req, err := http.NewRequest(method, u.String(), getReader())
 	if err != nil {
 		log.Debug("unable to create new request")
 		return 0, nil, err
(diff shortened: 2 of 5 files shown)


folbricht:
Sorry, already merged the logging PR which created a conflict on this one. Could you rebase it?

Kalmalyzer:
Rebase done!

README.md
```diff
@@ -238,6 +238,7 @@ Available configuration values:
 - `store-options` - Allows customization of chunk and index stores, for example comression settings, timeouts, retry behavior and keys. Not all options are applicable to every store, some of these like `timeout` are ignored for local stores. Some of these options, such as the client certificates are overwritten with any values set in the command line. Note that the store location used in the command line needs to match the key under `store-options` exactly for these options to be used. Watch out for trailing `/` in URLs.
   - `timeout` - Time limit for chunk read or write operation in nanoseconds. Default: 1 minute. If set to a negative value, timeout is infinite.
   - `error-retry` - Number of times to retry failed chunk requests. Default: 0.
+  - `error-retry-base-interval` - Number of seconds to wait before first retry attempt. Retry attempt number N for the same request will wait N times this interval. Default: 1 second.
```
folbricht, line 241: Wouldn't 0 be a better default? To match the prior behavior.

remotehttp_test.go
```diff
@@ -40,3 +43,259 @@ func TestHTTPStoreURL(t *testing.T) {
...
+		"permanent 503": {ChunkID{0, 0, 0, 5}, false, true, 5},
+		// HTTP 503 Bad Gateway - should retry, and a subsequent successful call should return that the chunk exists
+		"temporary 503, then chunk exists": {ChunkID{0, 0, 0, 6}, true, false, 2},
+		// HTTP 503 Bad Gateway - should retry, and a subsequent successful call should report that the chunk does not exist immediately
+		"temporary 503, then chunk does not exist": {ChunkID{0, 0, 0, 7}, false, false, 3},
+	}
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			u.Path = "/"
+			s, err := NewRemoteHTTPStore(u, StoreOptions{ErrorRetry: 5, ErrorRetryBaseInterval: time.Duration(1) * time.Microsecond})
```
folbricht, line 109: Don't really need the `time.Duration(1)` bit here, just `time.Microsecond` will be enough

remotehttp_test.go
```diff
@@ -40,3 +43,259 @@ func TestHTTPStoreURL(t *testing.T) {
...
+		"permanent 503": {ChunkID{0, 0, 0, 5}, "", true, 5},
+		// HTTP 503 Bad Gateway - should retry, and a subsequent successful call should return a successful chunk
+		"temporary 503, then chunk exists": {ChunkID{0x65, 0xa1, 0x28, 0xd0, 0x65, 0x8c, 0x4c, 0xf0, 0x94, 0x17, 0x71, 0xc7, 0x09, 0x0f, 0xea, 0x6d, 0x9c, 0x6f, 0x98, 0x18, 0x10, 0x65, 0x9c, 0x24, 0xc9, 0x1b, 0xa2, 0x3e, 0xdd, 0x71, 0x57, 0x4b}, "Chunk Content String 6", false, 2},
+		// HTTP 503 Bad Gateway - should retry, and a subsequent successful call should report that the chunk does not exist, thereby failing immediately
+		"temporary 503, then chunk does not exist": {ChunkID{0, 0, 0, 7}, "", true, 3},
+	}
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			u.Path = "/"
+			s, err := NewRemoteHTTPStore(u, StoreOptions{ErrorRetry: 5, ErrorRetryBaseInterval: time.Duration(1) * time.Microsecond, Uncompressed: true})
```
folbricht, line 195: same as above

remotehttp_test.go
```diff
@@ -40,3 +43,259 @@ func TestHTTPStoreURL(t *testing.T) {
...
+		"forbidden": {ChunkID{0, 0, 0, 4}, "4", "", true, 1},
+		// HTTP 503 Bad Gateway - should retry, but ultimately fail
+		"permanent 503": {ChunkID{0, 0, 0, 5}, "5", "", true, 5},
+		// HTTP 503 Bad Gateway - should retry, and a subsequent successful call should make the entire operation succeed
+		"temporary 503, then store chunk successful": {ChunkID{0x65, 0xa1, 0x28, 0xd0, 0x65, 0x8c, 0x4c, 0xf0, 0x94, 0x17, 0x71, 0xc7, 0x09, 0x0f, 0xea, 0x6d, 0x9c, 0x6f, 0x98, 0x18, 0x10, 0x65, 0x9c, 0x24, 0xc9, 0x1b, 0xa2, 0x3e, 0xdd, 0x71, 0x57, 0x4b}, "Chunk Content String 6", "Chunk Content String 6", false, 2},
+	}
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			u.Path = "/"
+			s, err := NewRemoteHTTPStore(u, StoreOptions{ErrorRetry: 5, ErrorRetryBaseInterval: time.Duration(1) * time.Microsecond, Uncompressed: true})
```
folbricht, line 283: same

folbricht:
Looking very good. Much cleaner and easier to read since you removed some of the duplication that had built-up over time. The only question is what to use as default for the retry interval.

README.md
Kalmalyzer: 0 is better for prior users, 1 is better for new users I'd say. Let's go with 0.

remotehttp_test.go
Kalmalyzer: Thanks, will fix.

Kalmalyzer:
I have implemented your suggested changes. Care to accept PR?

folbricht:
I can merge it tonight. There's still an issue with the units of the base interval but I can fix that right after merging if you don't get to it before. The issue is that right now it'll take those as nanoseconds, not seconds. But if you were to add `time.Second` to the calculation it would make the tests sleep for a long time. I suspect we just have to be specific in the readme that the config expects nanoseconds, similar to the other `timeout` value

remotehttp.go
```diff
@@ -95,94 +97,105 @@ func (r *RemoteHTTPBase) String() string {
...
+	if (err != nil) || (statusCode >= 500 && statusCode < 600) {
+		if attempt >= r.opt.ErrorRetry {
+			log.WithField("attempt", attempt).Debug("failed, giving up")
+			return 0, nil, err
+		} else {
+			log.WithField("attempt", attempt).WithField("delay", attempt).Debug("waiting, then retrying")
+			baseInterval := r.opt.ErrorRetryBaseInterval
+			if baseInterval == 0 {
+				baseInterval = time.Duration(0)
+			}
```
folbricht, line 164: no need for the condition, but won't hurt either. Will remove it at some point, depending on what solution we'll go with for the units of base-interval

folbricht:
Thanks for this PR. I made a small tweak to the docs to say that the interval is in nanoseconds in https://github.com/folbricht/desync/commit/20f3fba06cb754302f5d4281f1dd20cdca956bfb

Kalmalyzer:
Excellent, thanks!

PR #162, opened 2020-04-25
Archives always contain forward-slash paths, even on Windows

Kalmalyzer:
Some of the archive tests fail on Windows. Rather than using `filepath.Join` to construct within-archive paths (which will create paths like `dir1\dir2` on Windows, and `dir1/dir2` on Unix), those tests should use plain `path.Join` (which will create paths like `dir1/dir2` on all OSes). Fixed in this PR; now all tests pass on Windows.

### DIFF
--- a/archive_test.go
+++ b/archive_test.go
@@ -2,7 +2,7 @@ package desync
 
 import (
 	"os"
-	"path/filepath"
+	"path"
 	"reflect"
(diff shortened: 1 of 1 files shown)


folbricht:
Thanks for the fix. Not sure why I didn't see that fail before.

PR #163, opened 2020-04-25
GitHub actions validation

Kalmalyzer:
I have set up minimal CI via GitHub Actions. As you can see, there is currently a problem with the tests on Windows. #162 fixes that problem with the tests.

I'm not sure whether or not you are interested in having this form of validation, but I find it handy. Particularly in that you will automatically see these test results in-line in peoples' pull requests in the future.

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -1,0 +1,38 @@
+
+name: Validate
+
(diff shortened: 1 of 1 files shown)


folbricht:
Good idea, didn't think of that.

PR #164, opened 2020-04-28
GitHub validation improvements

Kalmalyzer:
Couple of smaller fixes for the GH Actions validation logic:

* The validation is now a matrix job. No more separate Windows/Linux flows.
* Validation will also run on MacOSX
* The set of installed Go modules are cached between runs. The cache is invalidated whenever the set of modules change. This makes building a few seconds faster. The big time sink is still download & installation of go itself; 20 seconds on Linux, 30 seconds on MacOSX, 50 seconds on Windows.

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -12,10 +12,14 @@ on:
       - '**'
 
 jobs:
-  validate-linux:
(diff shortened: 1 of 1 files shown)


folbricht:
Thank you

PR #166, opened 2020-05-02
The GCS index store should use the path without trailing filename as its root

Kalmalyzer:
Attempting to use an URI like `gs://bucket/folder/filename.caibx` used to result in an actual retrieval path of `gs://bucket/folder/filename.caibx/filename.caibx`. This change fixes it. The GCS index store now behaves like other stores.


folbricht:
Thanks for the fix

PR #167, opened 2020-05-04
Rewritten gcs filename prefix logic to never include leading slash

Kalmalyzer:
The original logic in `gcs.go` would create prefix strings with leading slashes. This would result in item names like `/file1.txt`, `/folder1/file1.txt`, etc.

These paths should not have had leading slashes. For files within folders, Google Cloud Storage ignored the leading slash, but for files in the root, GCS would create a nameless folder - i.e. a folder with the name "". This caused problems for index stores when placing files into the root of a bucket.

This change ensures that the path prefix for the root folder is "" and for an actual folder is `folder1/`, etc.

### DIFF
--- a/gcs.go
+++ b/gcs.go
@@ -32,6 +32,23 @@ type GCStore struct {
 	GCStoreBase
 }
 
+// normalizeGCPrefix converts path to a regular format,
+// where there never is a leading slash,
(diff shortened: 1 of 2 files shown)


folbricht:
Thank you

PR #170, opened 2020-08-16
Use semantic import path for minio so we always get version 6

srd424:
This seems to fix build failures seen when running in GOPATH mode, and seems to be [required anyway](https://golang.org/cmd/go/#hdr-Module_compatibility_and_semantic_versioning) (disclaimer - I'm not a go person!)


folbricht:
Thanks for that

PR #173, opened 2020-10-22
bitmap and SIGHUP handler for writing state

jbeguin:
hello,
I'm Jean-Marie from wisper.
I replace state file with Bitmap from https://godoc.org/github.com/boljen/go-bitmap
I also add signal.Notify SIGHUP for writing state in sparse-file.go with a new chan because of accessibility between loader and sighup
Hope you can have a look.

### DIFF
--- a/sparse-file.go
+++ b/sparse-file.go
@@ -5,8 +5,12 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"os/signal"
 	"sort"
 	"sync"
+	"syscall"
+
+	"github.com/boljen/go-bitmap"
 )
 
 // SparseFile represents a file that is written as it is read (Copy-on-read). It is
@@ -62,6 +66,15 @@ func NewSparseFile(name string, idx Index, s Store) (*SparseFile, error) {
 		}
 	}
 
+	// save state file on SIGHUP.
+	sighup := make(chan os.Signal)
+	signal.Notify(sighup, syscall.SIGHUP)
+	go func() {
+		for range sighup {
+			loader.writeState()
+		}
+	}()
+
 	return &SparseFile{
 		name:   name,
 		idx:    idx,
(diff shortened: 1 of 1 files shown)


sparse-file.go
```diff
@@ -214,29 +223,36 @@ func (l *sparseFileLoader) loadChunk(i int) error {
...
 	defer l.mu.Unlock()
-	b := make([]byte, 0, len(l.done))
-	for _, done := range l.done {
-		if done {
-			b = append(b, '1')
-		} else {
-			b = append(b, '0')
-		}
+
+	f, err := os.OpenFile(l.name+".state", os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0755)
```
folbricht, line 240: Wouldn't it be easier to just use `ioutil.WriteFile()` here instead, since it doesn't use the io.Reader anymore? No need for explicit calls to Write or Sync

folbricht:
Thanks for that. Was considering implementing implementing this myself directly in desync but this is better.

jbeguin:
I must said I'm new in go and don't know all best practices, libs and tips.
I'll check that.
thanks

folbricht:
No worries. So `os.OpenFile()` is a low-level function that you need when you want specific perms or flags. The next one higher up is `os.Create()` which is great if you want an io.Writer/Reader. But if you already have a byte slice and want to just write it into a file, `ioutil.WriteFile` will be fine. All of those build on each other.

sparse-file.go
```diff
@@ -214,30 +223,21 @@ func (l *sparseFileLoader) loadChunk(i int) error {
...
-		if done {
-			b = append(b, '1')
-		} else {
-			b = append(b, '0')
-		}
-	}
-	_, err := w.Write(b)
-	return err
+
+	return ioutil.WriteFile(l.name+".state", l.done.Data(false), 0755)
```
folbricht, line 240: Sorry, one more, I think this should be be 0644 since it's not executable

jbeguin: ok ! sorry and thanks a lot
copy/paste is not our friend :)

folbricht (approved):

folbricht:
Thank you

jbeguin:
no, thank you :)

PR #174, opened 2020-10-25
Contrib

jbeguin:
hello,
sorry for this new pr.
I miss the basic check when loading state file.
I think it should work like this.

### DIFF
--- a/sparse-file.go
+++ b/sparse-file.go
@@ -5,8 +5,12 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"os/signal"
 	"sort"
 	"sync"
+	"syscall"
+
+	"github.com/boljen/go-bitmap"
 )
 
 // SparseFile represents a file that is written as it is read (Copy-on-read). It is
@@ -62,6 +66,15 @@ func NewSparseFile(name string, idx Index, s Store) (*SparseFile, error) {
 		}
 	}
 
+	// save state file on SIGHUP.
+	sighup := make(chan os.Signal)
+	signal.Notify(sighup, syscall.SIGHUP)
+	go func() {
+		for range sighup {
+			loader.writeState()
+		}
+	}()
+
 	return &SparseFile{
 		name:   name,
 		idx:    idx,
@@ -86,12 +99,7 @@ func (sf *SparseFile) Length() int64 {
 // Close saves the state of file, basically which chunks were loaded
 // and which ones weren't.
 func (sf *SparseFile) Close() error {
-	f, err := os.Create(sf.name + ".state")
-	if err != nil {
-		return err
-	}
-	defer f.Close()
-	return sf.loader.writeState(f)
+	return sf.loader.writeState()
 }
 
 // ReadAt reads from the sparse file. All accessed ranges are first written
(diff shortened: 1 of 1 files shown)


folbricht:
Thank you. Missed that one.

I'm about to rework a lot of this code to add pre-loading of chunks on startup. You might want to create a new branch when I'm done

jbeguin:
hello,
for some purposes only on sparse mount branch, but maybe useful for all.
my branch may be diverge a bit from yours.
I add a file logger with log level option, that can be helpful with
non-local stores.
also, a file which record progress from preload state.
and finally a telnet server binding on localhost considered to be used
through socat like, that mean previous log files could be removed.
would you in interest ?
I don't know if the philosophy of your code is respected with singleton
monitor, but I know that you I have great look on code (since I'm 1 month
older in go)
also, I start thinking encryption of chunks, and I see the problem of
read/write about calls of Compress/Uncompress.
How do you see that point ?
I have some ideas, not already coded.;
maybe, just add a flag read/write when Compress/Uncompress, then
encrypt/desencrypt.
hope we can merge all of these branches, but maybe not necessary for you.
jean-marie


Le dim. 25 oct. 2020 à 15:53, Frank Olbricht <notifications@github.com> a
écrit :

folbricht:
The issue with encryption is really my sub-optimal implementation of compressed/uncompressed stores. It really should have been done with modifiers that are layered on top of the raw storage, some to decompress/compress and encrypt/decrypt on top of that. I don't really have time at the moment to go through everything and clean that up.

There are also some optimizations (in the chunk-server) around avoiding additional compress/decompress operations that would be impacted I suspect. Compress/decompress likely needs to be re-implemented in a cleaner way before adding encryption. Unfortunately I don't have enough time to spend on this at the moment.

folbricht:
It's quite hard to review without a PR or squashing the commits, but I do see some of the complexity I'd rather avoid. Passing the key around and then adding conditions in many places. Especially the extra edge cases this creates with the uncompressed flag aren't ideal.
You can join https://gitter.im/desync-casync-client/Lobby and then ping me directly, it might be a better way to communicate than this ticket. Let's see if we can find a better way, I may have some time the next couple weeks.

jbeguin:
hard to do un-compress, even then false value as default.
compress or enrcypt is touchy, for parameters.
ok. false is default. how do you that ?
then from inner packaged(dst_comp, dst_encrypt) is the better for me.
i am not good in  english, so i al glad to your chanel.
probably join it, but i'm just in tipe in english

Le dim. 20 déc. 2020 à 18:21, Frank Olbricht <notifications@github.com> a
écrit :

jbeguin:
also, there log file and local telnet monitoring on my branch . hard way to
merge, or not

jbeguin:
i mus finish .sh to do full test adn bench. let me some days :)

jbeguin:
the idea from gived key each funtion is nobidy knokn there.

jbeguin:
my  father, lot years ago.
the man with mandoline and loud voice
https://duckduckgo.com/?q=lyac%27h&atb=v165-2__&iar=videos&iax=videos&ia=videos&iai=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DYsPiAtSJfHM
totallly drunk, see you soon

Le dim. 20 déc. 2020 à 18:21, Frank Olbricht <notifications@github.com> a
écrit :

Issue #175, opened 2020-10-28
S3 transient tcp errors on large uploads


Joe-Improbable:
I've encountered a transient issue when performing large uploads to an S3 store. 

I occasionally run into this error:
```
Put "<s3 bucket name>/<chunk id>/<chunk id>.cacnk": write tcp <ip>-><ip>: use of closed network connection
```

A simple solution to this problem would just be to retry any failed PUT (or GET) requests, similar to how the `RemoteHTTPStore` does with `error-retry` option. 

Minio-go implements retries but only for errors it deems retry-able  ([listed here](https://github.com/minio/minio-go/blob/master/retry.go#L90)), unfortunately any TCP errors won't be retried even though a number of them can be solved by retrying the request.

PR #176, opened 2020-10-28
add retries to S3 GetChunk and PutChunk

Joe-Improbable:
This PR should solve (or at least significantly reduce the chance of) #175

It changes the S3 store to also respect the `error-retry` option and retry any failed GET or PUT requests


folbricht (approved):

folbricht:
Thank you. I'm wondering if longer-term it'd be better to use a wrapper that implements the Store interface to do the retries, rather than having it in every store.

PR #177, opened 2020-11-04
Local filesystem Prune clears up partially downloaded chunks

Joe-Improbable:
This PR makes any partially downloaded chunk (in the form `.tmp.canck.*`) be removed when `Prune()` is called on a local store.


folbricht (approved):

folbricht:
Thank you

PR #181, opened 2020-12-26
Refactoring to support storage modifiers

folbricht:
This change abstract the handling of uncompressed chunk storage with the goal of supporting other storage modifier layers later, namely for encryption.


coverter.go
```diff
@@ -0,0 +1,97 @@
+package desync
```
charles-dyfis-net, line 1: Is there a typo in this filename?

charles-dyfis-net (approved):
This is elegant, well-tested and well-commented. If you were to merge this (as the groundwork for adding encryption) as a separate PR from the actual crypto implementation, I could see merging it as-is as a code-quality enhancement (albeit maybe retitled).

coverter.go
folbricht: good catch. Thanks, Fixed

folbricht:
I do actually have most of the remaining encryption code ready to go on top of this, with a few more modifications to the large refactoring (just missing a couple tests and docs). You're right, should probably use this PR for the refactoring only and add the encryption layer in another one.

PR #188, opened 2021-04-22
datadog is now DataDog, enforced by their go.mod

charles-dyfis-net:
On February 11th, datadog/zstd introduced a `go.mod` describing the library as being `DataDog/zstd`. This introduced build breakage.

Quoting their commit message (DataDog/zstd@9847581ee76d182dd6ad4a0e23795f09297ba123):

> Since we have no external dependancies, the module file is pretty barebone.
> 
> "DataDog" is capitalized as is because it is the current github URL, even though github URLs are case-insensitive; talking to GitHub support, it looks like a migration to a different casing (i.e: datadog) is not 100% compatible (i.e: some systems that are case-sensitive for project names could be impact).
>  
> Since we always run tests for go version N & N-1 (N being latest), we currently support fully 1.14 as a minimum version.

### DIFF
--- a/compress_datadog.go
+++ b/compress_datadog.go
@@ -3,7 +3,7 @@
 package desync
 
 import (
-	"github.com/datadog/zstd"
+	"github.com/DataDog/zstd"
 )
 
 // Compress a block using the only (currently) supported algorithm
--- a/go.mod
+++ b/go.mod
@@ -4,8 +4,8 @@ go 1.15
 
 require (
 	cloud.google.com/go/storage v1.12.0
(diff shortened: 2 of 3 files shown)


charles-dyfis-net:
FYI, this was prompted by a report of the following:

`go get: github.com/datadog/zstd@v1.4.5 updating to github.com/datadog/zstd@1.4.8: parsing go.mod: module declares its path as: github.com/DataDog/zstd but was required as: github.com/datadog/zstd`

folbricht (approved):

folbricht:
Thank you.

PR #191, opened 2021-05-07
Don't fail when no S3 credentials file exists

### LINKED ISSUE #190: ~/.aws/credentials now mandatory even pulling from publicly-readable source?
I have a user report of a `failed to load shared credentials file: open /home/<username>/.aws/credentials: no such file or directory` from a user trying to download from a publicly-shared `s3+https://...` URI (that, being public, should not require any credentials), resolved by downgrading to an older version of desync.

folbricht:
Fixes #190 

desync should be able to access public buckets and not require a credentials file to be present.


charles-dyfis-net (approved):
This change has been tested to work in practice to address #190.

Issue #196, opened 2021-08-29
cache command should support --seed parameter


om26er:
The cache command currently downloads the whole set of chunks for an index. However I want only the chunks that are missing as compared to my seed. The `desync cache` command should optionally take a `--seed` parameter. Here is my use case

I want to keep two versions of a software (lets call them version 1 and version 2) that is a SquashFS container. I want to keep the full file of version 2 on the system and only keep the chunks of version 1 that are different from version 2. Using this technique, I can implement a "client side revert" feature and maintaining only the delta of the old version and hence saving lots of space.

folbricht:
If I understand it correctly, this seems like what `desync chop --ignore <index>` does currently, just for the `cache` subcommand. If that's the case, it should be straightforward to add `--ignore`

om26er:
Just played with the `chop` command and yeah, we need a similar treatment to the `cache` command and this should work great.

folbricht:
Please try out the `issue-196` branch. It should do what you want.

folbricht:
Merged in https://github.com/folbricht/desync/pull/197, let me know if you run into issues.

PR #197, opened 2021-08-29
Add --ignore and --ignore-chunks options to cache command

folbricht:
Implements #196


om26er:
Gave it a try, works as expected

PR #199, opened 2021-11-25
Add information about the seed indexes.

RyuzakiKK:
This adds the ability to add one or more seed indexes to the `info`
command.

The resulting output will contain the number of chunks, from the index
provided in input, that are available in the seed. Similarly to what
it's already done with the store.

Additionally it also shows the size of the unique (non duplicated)
chunks that are not available from the seed.
This information can be used as a way to roughly estimate the download
size of a particular update. For example if we use:
```
desync info --seed local.caibx -s https://[...]/store update.caibx
```
We will get in output the size of the chunks that are required by
`update.caibx` but that are not already available in our local seed. So
these will be the chunks that we must download from the remote store.

### DIFF
--- a/cmd/desync/info.go
+++ b/cmd/desync/info.go
@@ -13,6 +13,7 @@ import (
 type infoOptions struct {
 	cmdStoreOptions
 	stores      []string
+	seeds       []string
 	printFormat string
 }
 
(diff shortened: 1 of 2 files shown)


charles-dyfis-net (approved):
LGTM on a code read, and it's certainly useful functionality.

folbricht (approved):
Thank you

PR #200, opened 2021-11-25
Add --repair option for command that work with cache

tanshihaj:
Chunks in local cache can corrupt for many reasons and by default `desync` fails when met them. They can be removed by running `desync verify --repair --store /path/to/cache` and runnnig `desync untar ...` again, however it is quite a long operation.
It could be convenient to repair them by redownloading invalid chunks from store when we met them during `desync untar` and other command which work with cache.

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -32,7 +32,7 @@ func (c Cache) GetChunk(id ChunkID) (*Chunk, error) {
 	default:
 		return chunk, err
 	}
-	// At this point we failed to find it in the local cache. Ask the remote
+	// At this point we failed to find chunk in the local cache. Ask the remote
 	chunk, err = c.s.GetChunk(id)
 	if err != nil {
 		return chunk, err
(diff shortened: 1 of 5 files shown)


folbricht:
Thanks for the PR. I do like the feature, though not a fan of passing the repair flag into so many places. You should be able to simplify that quite a bit.

If you look at https://github.com/folbricht/desync/blob/master/cmd/desync/store.go#L22, it gets `cmdStoreOptions` passed in. So if you added your flag [here](https://github.com/folbricht/desync/blob/master/cmd/desync/options.go#L12), perhaps calling it `--cache-repair` or similar to be more specific, then you could either pass the boolean to `NewCache()` [here](https://github.com/folbricht/desync/blob/master/cmd/desync/store.go#L40).

To avoid passing a flag to `NewCache()`, one could create a new object that implements the `WriteStore` interface and layer that on top of the store in https://github.com/folbricht/desync/blob/master/cmd/desync/store.go#L40. That way there wouldn't need to be any boolean parameter at all, though it would require implementing a store layer whose only job it is to turn InvalidChunk into ChunkNotFound basically.

folbricht (approved):
Thanks for the update.

PR #202, opened 2021-12-07
Added --error-retry store option to retry network errors

tanshihaj:
Currently when downloading chunks from s3 store over unreliable network, TCP session can fall apart and server will begin to send `RST` package. And because of that commands that work with s3 store fails. That would be great to have and option to retry download in case of network error.

We already have `ErrorRetry` entry in `StoreOptions`, in this commit I'm adding `--error-retry` cli option that overrides `ErrorRetry` from configs. Also I think that there is bug with retry here https://github.com/folbricht/desync/blob/master/s3.go#L97-L99: `GetObject()` will return error only when object or bucket name is invalid and retry will no fix it. At the same time we ignore network errors that can raise during `ioutil.ReadAll` https://github.com/folbricht/desync/blob/master/s3.go#L104 that can be fixed with retry.

### DIFF
--- a/s3.go
+++ b/s3.go
@@ -71,7 +71,7 @@ func (s S3StoreBase) String() string {
 	return s.Location
 }
 
-// Close the S3 base store. NOP opertation but needed to implement the store interface.
+// Close the S3 base store. NOP operation but needed to implement the store interface.
 func (s S3StoreBase) Close() error { return nil }
 
 // NewS3Store creates a chunk store with S3 backing. The URL
@@ -94,7 +94,7 @@ retry:
 	attempt++
 	obj, err := s.client.GetObject(s.bucket, name, minio.GetObjectOptions{})
 	if err != nil {
-		if attempt < s.opt.ErrorRetry {
+		if attempt <= s.opt.ErrorRetry {
 			goto retry
 		}
 		return nil, errors.Wrap(err, s.String())
@@ -102,17 +102,20 @@ retry:
 	defer obj.Close()
 
 	b, err := ioutil.ReadAll(obj)
-	if e, ok := err.(minio.ErrorResponse); ok {
-		switch e.Code {
-		case "NoSuchBucket":
-			err = fmt.Errorf("bucket '%s' does not exist", s.bucket)
-		case "NoSuchKey":
-			err = ChunkMissing{ID: id}
-		default: // Without ListBucket perms in AWS, we get Permission Denied for a missing chunk, not 404
-			err = errors.Wrap(err, fmt.Sprintf("chunk %s could not be retrieved from s3 store", id))
-		}
-	}
 	if err != nil {
+		if attempt <= s.opt.ErrorRetry {
+			goto retry
+		}
+		if e, ok := err.(minio.ErrorResponse); ok {
+			switch e.Code {
+			case "NoSuchBucket":
+				err = fmt.Errorf("bucket '%s' does not exist", s.bucket)
+			case "NoSuchKey":
+				err = ChunkMissing{ID: id}
+			default: // Without ListBucket perms in AWS, we get Permission Denied for a missing chunk, not 404
+				err = errors.Wrap(err, fmt.Sprintf("chunk %s could not be retrieved from s3 store", id))
+			}
+		}
 		return nil, err
 	}
 	return NewChunkFromStorage(id, b, s.converters, s.opt.SkipVerify)
--- a/s3_test.go
+++ b/s3_test.go
@@ -1,0 +1,258 @@
+package desync
+
+import (
+	"bufio"
+	"context"
+	"errors"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+	"testing"
+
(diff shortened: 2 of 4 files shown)


tanshihaj:
Maybe it is also good idea to set some non-zero default value (about 3) for `--error-retry`

s3.go
```diff
@@ -94,14 +95,12 @@ retry:
 	attempt++
 	obj, err := s.client.GetObject(s.bucket, name, minio.GetObjectOptions{})
 	if err != nil {
-		if attempt < s.opt.ErrorRetry {
-			goto retry
-		}
 		return nil, errors.Wrap(err, s.String())
```
folbricht, line 98: I believe the retry here is still necessary, since it's still going to make a network call to lookup the object. That initial call could fail as well

s3.go
```diff
@@ -111,6 +110,10 @@ retry:
 		default: // Without ListBucket perms in AWS, we get Permission Denied for a missing chunk, not 404
 			err = errors.Wrap(err, fmt.Sprintf("chunk %s could not be retrieved from s3 store", id))
 		}
+	} else if errors.As(err, &opError) {
```
folbricht, line 113: Why not check for `net.Error`, perhaps in a type-switch? OpError doesn't cover all possible network-related problems

s3_test.go
```diff
@@ -0,0 +1,244 @@
...
+	file, err := os.Open(filePath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			resp := response(request, http.Header{}, 404, "")
+			resp.Write(conn)
+		} else {
+			resp := response(request, http.Header{}, 500, err.Error())
+			resp.Write(conn)
+		}
+	}
```
folbricht, line 59: Shouldn't there be a `return` here or a `t.Fatal`? Without that the `file.Stat` is going to panic if file is nil (like in case of failure). Also, it's only a test, but may be worth doing a `defer file.Close()` here

s3_test.go
```diff
@@ -0,0 +1,244 @@
...
+
+	stat, err := file.Stat()
+	if err != nil {
+		resp := response(request, http.Header{}, 500, err.Error())
+		resp.Write(conn)
```
folbricht, line 64: Same here, it'll panic below if `stat` is nil

s3_test.go
```diff
@@ -0,0 +1,244 @@
...
+
+// Run S3 server that can respond objects from `store`
+// if `errorTimesLimit` > 0 server will send RST packet `errorTimesLimit` times after sending half of file
+func getTcpS3Server(t *testing.T, ctx context.Context, bucket, store string, errorTimesLimit int) (net.Listener, *errgroup.Group) {
+	group := errgroup.Group{}
+	var errorTimes int
+	// using localhost + resolver let us work on hosts that support only ipv6 or only ipv4
+	ip, err := net.DefaultResolver.LookupIP(ctx, "ip", "localhost")
+	if err != nil || len(ip) < 1 {
+		t.Fatal(err)
```
folbricht, line 113: Could be a slightly confusing test failure if err=nil, but there's no IP

s3_test.go
```diff
@@ -0,0 +1,244 @@
...
+				return err
+			}
+			reader := bufio.NewReader(conn)
+			request, err := http.ReadRequest(reader)
+			if err != nil {
+				return err
+			}
+
+			matches := objectGetMatcher.FindStringSubmatch(request.URL.Path)
+			// t.Logf("shit is %q | %q | %v | %s", request.Method, request.URL, request.Header, matches)
```
folbricht, line 143: Leftover from debugging?

s3_test.go
```diff
@@ -0,0 +1,244 @@
...
+			if err != nil {
+				if errors.Is(ctx.Err(), context.Canceled) {
+					return nil
+				}
+				return err
+			}
+			reader := bufio.NewReader(conn)
+			request, err := http.ReadRequest(reader)
+			if err != nil {
+				return err
```
folbricht, line 139: Technically, there should be a `conn.Close()` here since one was opened above

s3_test.go
tanshihaj: Yep, indeed) Will remove

s3.go
tanshihaj: No it will not work since `net.Error` is interface - so any error that have `Error() string`, `Timeout() bool` and `Temporary() bool` will cast to `net.Error` which is not correct. I know that `net.OpError` cover all network errors, but from my little experience it cover most of errors that can be fixed by retry. But if you think that retrying on every network error is ok (even if retry will not fix error) we can just retry on any error here.

s3.go
tanshihaj: According to code `GetObject()` return error only here https://github.com/minio/minio-go/blob/master/api-get-object.go#L37 and here https://github.com/minio/minio-go/blob/master/api-get-object.go#L40. But not I think this behavior can change I better bring back this lines.

s3_test.go
tanshihaj: Yes, you are right, will return here and close file

folbricht (approved):
Thank you

Issue #117, opened 2019-08-12
Redesign seed sequencer


folbricht:
When running `extract` with seeds, there are a couple of issues that seem different but happen in the same piece of code. Fixing those requires a rewrite of the code that determines what seed should be used.

1. The current algorithm tries to find the seed with the longest sequence of matching chunks, regardless if that sequence is aligned and can be reflinked or not. Ideally, the algorithm should always use "clone-able" ranges first and if there aren't any more, then also consider "copy-able". This will result in optimal storage efficiency and performance.

2. When a seed is corrupt (the hash of a chunk doesn't match), the whole process fails. It'd be much better to either evict the whole seed and carry on without it, or to keep going and just get the damaged chunks by other means (other seeds or chunk store).

Getting both of those right and doing it all concurrently may require additional complexity, perhaps the self-seed, where chunks from earlier sections of the target file are copied/cloned to later sections, may need to be dropped and be replaced with a static execution plan calculated beforehand, not during execution.

PR #203, opened 2021-12-07
Add option to discard invalid seeds

RyuzakiKK:
If we have a seed index for a file/device that is corrupted we should
try harder to continue anyway by discarding the invalid seed and
fallback to the potentially other seeds and/or by just using the store.

Partially addresses #117

----

At the beginning I tried to just mark the unexpected chunk as invalid, but realigning the "good" chunks to compensate the corrupted values required too much work and was probably not worth it.

Please let me know if you'd prefer to have a launch option to control this behavior or if it's okay to just always do it.


RyuzakiKK:
While thinking more about that I think that having an option to switch between "try to continue without the seed" and "bail out" might be useful, and it also gives us the ability to expand it later to additionally more clever things instead of just entirely discarding the seed.

RyuzakiKK:
Now by default I left the current behavior of failing on invalid seeds and added a new option `--skip-invalid-seeds` to discard them and continue.

I created a type `InvalidSeedAction` instead of a single boolean to make it easier in the future to add additional actions that we might want to add.

folbricht (commented):
Thank you. I'll need a bit more time to review this since it's in a rather complex part of the code base, but I should be able to get to it later this week

assemble.go
```diff
@@ -8,6 +8,34 @@ import (
...
+	isaBailOut InvalidSeedAction = iota
+	isaSkip
+)
+
+type AssembleOptions struct {
+	n   int
+	isa InvalidSeedAction
+}
+
+func NewAssembleOptions(n int, skipInvalid bool) AssembleOptions {
```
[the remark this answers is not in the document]
RyuzakiKK: Sorry for the possible obvious question but I'm new to Golang.
Are you suggesting to export the fields of `AssembleOptions`, by renaming them with a capital letter, or something else?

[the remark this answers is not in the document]
folbricht: Nope, that's exactly it. It's a common patters, keep all the options in a struct, with reasonable defaults. Having a constructor can sometimes help, but here all it does is set the same values. If you export the fields then the caller can simply pass a populated struct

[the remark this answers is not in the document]
RyuzakiKK: Thanks for the explanation. With the latest revision I removed the constructor.

cmd/desync/extract_test.go
```diff
@@ -96,3 +100,34 @@ func TestExtractWithFailover(t *testing.T) {
...
+			cmd := newExtractCommand(context.Background())
+			cmd.SetArgs(append(test.args, test.output))
+
+			// Redirect the command's output and run it
+			stderr = ioutil.Discard
+			cmd.SetOutput(ioutil.Discard)
+			_, err := cmd.ExecuteC()
+			require.Error(t, err)
+		})
+	}
```
folbricht, line 132: This currently depends on desync detecting any errors in the assembly step. Given the complexities of dealing with corrupted seeds, do you think it'd make sense to validate the output file? Like perhaps read it back from disk and calculating the SHA256 of it, then comparing to the expected value.

assemble.go
```diff
@@ -8,6 +8,21 @@ import (
...
+type InvalidSeedAction int
+
+const (
+	IsaBailOut InvalidSeedAction = iota
+	IsaSkip
+)
+
+type AssembleOptions struct {
+	N   int
+	Isa InvalidSeedAction
```
folbricht, line 23: Feel free to rename `Isa` to `InvalidSeedAction` (it won't clash with the type above) as well if that makes things more readable. If you prefer to keep it short, `ISA` would work too as per https://github.com/golang/go/wiki/CodeReviewComments#initialisms

fileseed.go
```diff
@@ -22,28 +23,32 @@ func NewIndexSeed(dstFile string, srcFile string, index Index) (*FileSeed, error
...
 		s.pos[c.ID] = append(s.pos[c.ID], i)
 	}
 	return &s, nil
 }
 
-// LongestMatchWith returns the longest sequence of of chunks anywhere in Source
-// that match b starting at b[0]. If there is no match, it returns nil
-func (s *FileSeed) LongestMatchWith(chunks []IndexChunk) (int, SeedSegment) {
+// LongestMatchWith returns the longest sequence of chunks anywhere in Source
+// that match b starting at b[0]. If there is no match, it returns a length of zero
```
folbricht, line 35: Looks like my comments were wrong, talking about `b` instead of `chunks`. Would you mind fixing them while you're in here?

fileseed.go
```diff
@@ -52,7 +57,27 @@ func (s *FileSeed) LongestMatchWith(chunks []IndexChunk) (int, SeedSegment) {
...
+		src, err := os.Open(s.srcFile)
+		if err != nil {
+			return 0, nil, err
+		}
+		defer src.Close()
+		// We validate the segment here, there is no need to do it again when we write the data
+		segment = newFileSeedSegment(s.srcFile, match, s.canReflink, false)
+		if err = segment.validate(src); err != nil {
+			return 0, nil, err
+		}
```
folbricht, line 70: This will work, but I suspect it's going to have quite an impact on performance. That's because previously the validation (which requires open, read, hash, close) would only be done on the longest matches. With this change, it'll validate every possible seed even if it's not the longest and is never chosen to be used. It'll also do that many times over, even re-check chunks that were checked in previous iterations.

I believe, but haven't actually tried, that you can delay validation as before by making a change in seqencer.go instead. I'll leave a comment below.

sequencer.go
```diff
@@ -7,28 +7,46 @@ package desync
...
 // store. If done is true, the sequencer is complete.
-func (r *SeedSequencer) Next() (segment IndexSegment, source SeedSegment, done bool) {
+func (r *SeedSequencer) Next() (segment IndexSegment, source SeedSegment, done bool, err error) {
 	var (
 		max     uint64
 		advance = 1
 	)
 	for _, s := range r.seeds {
-		n, m := s.LongestMatchWith(r.index.Chunks[r.current:])
+		// If we already know that this seed is invalid, we just skip it
```
folbricht, line 33: You could probably break up this loop into 2 separate loops and get delayed validation, as well as chunk-eviction.
- One loop to collect the longest matches from every seed. Basically the upper part of the existing loop. You'd build a slice of candidates
- Below, one loop to iterate over the candidates, find the longest, perform validation if needed (perhaps you have to add `NeedsValidation()` to the `SeedSegment` interface). If the validation fails you mark the whole seed as invalid and move to the next candidate.

And if you want to go one step further, this split would even allow you to skip bad sections of seeds, while keeping using the seed. Could add that as a 3rd mode of operation, only skip bad candidates, but not invalidate the whole seed. It can be slow, but that's a tradeoff, perhaps network bandwidth is more important and trying to use a damaged seed makes sense in some use-cases.

folbricht (commented):
Great work. Please take a look at the performance issue and let me know what you think.

sequencer.go
RyuzakiKK: Thanks a lot for the review. I should be able to fix the implementation with your suggestions by tomorrow.

> And if you want to go one step further, this split would even allow you to skip bad sections of seeds, while keeping using the seed. Could add that as a 3rd mode of operation, only skip bad candidates, but not invalidate the whole seed. It can be slow, but that's a tradeoff, perhaps network bandwidth is more important and trying to use a damaged seed makes sense in some use-cases.

This is actually the first thing that I tried but then I decided to implement first the "discard the whole seed" option.
In my first attempt I marked the single chunks that were invalid and kept the rest as is.
While this was working fine for very simple cases, where the invalid file/image had just a few inverted bits, it kinda failed when instead entire chunks where missing or there were additional unexpected chunks.
The remaining chunk hashes were indeed correct, but they were pointing to the wrong blocks, so this required some sort of realigning I guess?

This is a use case that I'm actually quite interested in and wanted to explore more. This is because we have a situation where it's possible to modify a partition that is usually read-only (a "developer" mode that grants you RW permissions).
In this case we may want to push an update on top of an image that has additional files, not available in the index file `.caibx`, or even a few less.
And doing a `desync make` from scratch is quite an expensive operation, that should be avoided if possible.

fileseed.go
RyuzakiKK: Right, I implemented your suggestion of moving the validation out in `sequencer.go` so that we just do the validation for the longest match.

assemble.go
RyuzakiKK: Done, I used `InvalidSeedAction`.

cmd/desync/extract_test.go
RyuzakiKK: Sure, something similar to the `validate()` that it's currently in `fileseed.go`, but doing it against the output file.

I guess I can look into it. Would you prefer it to be in this PR or in a followup?

folbricht: In another PR is just fine. You could just hash the whole output file and compare it against a fixed value I suppose

index.go
```diff
@@ -24,10 +24,15 @@ type Index struct {
...
+// IsInvalid will be set to true.
 type IndexChunk struct {
-	ID    ChunkID
-	Start uint64
-	Size  uint64
+	ID        ChunkID
+	Start     uint64
+	Size      uint64
+	Validated bool
+	IsInvalid bool
```
folbricht, line 35: I don't think these two are required, are they? Those should only be needed on the seed, not the chunk-level

sequencer.go
folbricht: Thanks for making this change. You should be able to extend this quite easily to use as much of the seed as possible and not fail completely. You'd just need a 3rd case checking for something like `InvalidSeedActionContinue` below (with a switch-case).

In another PR of course.

sequencer.go
```diff
@@ -1,42 +1,96 @@
...
+	return len(m)
+}
+
+func (m matchingCandidateList) Less(i, j int) bool {
+	// Sort the candidates in descending order based on the segment size
+	return m[i].source.Size() > m[j].source.Size()
+}
+
+func (m matchingCandidateList) Swap(i, j int) {
+	m[i], m[j] = m[j], m[i]
```
folbricht, line 34: The sort-interface here actually opens up more possibilities. What desync should have done from the beginning is prefer reflink over simple copies when dealing with seeds. It's much more efficient (time and space) to reflink a chain of lets say 3 chunks, than it is to copy 5 chunks, even though that's larger.

You don't need to do that now, but it's something that could be added on top of your change. If you are on a filesystem that supports reflinks that may be something to consider.

folbricht:
Could you take a look at the two extra fields that were added to IndexChunk? I suspect they're leftover from an earlier implementation and can be removed. Other than that this looks good. Thanks for implementing this. It opens up more possibilities as well

cmd/desync/extract_test.go
RyuzakiKK: The expected hash for the whole file is it already available in the index file or should we calculate that as well?

index.go
folbricht: Hmm, I see. `IndexChunk` is a fundamental data structure that is used to describe chunks, and not related to seeds. I believe it's the wrong place to store these flags and could lead to all kinds of problems (including race conditions). It'd be better to store this somewhere else, like in the fileseed itself, or remove that for now to keep it simple. That would then require a re-validation if the same chunks are loaded again but that's basically what happened before too, so not slower than previously.
If you store it in the fileseed, basically wrapping the IndexChunk, you'll need to deal with concurrency and avoid a data race.

RyuzakiKK: Done in the latest revision.

sequencer.go
```diff
@@ -1,42 +1,94 @@
...
+
+		if n > 0 {
+			candidate = append(candidate, matchingCandidate{s, m, n})
+		}
+	}
+
+	// Sort the candidates from the longest to the shortest
+	sort.Sort(matchingCandidateList(candidate))
+	for _, matching := range candidate {
+		err := matching.source.Validate()
```
folbricht, line 70: Thinking about it a bit more I remember now why I didn't do it this way originally and delayed the validation. This code here runs in a single goroutine, while the copying of data happens concurrently. So previously we had N goroutines validating and copying the seed chunks. Now we validate in a single one, but still write concurrently. This should indeed be slower in that case. While I still like this change, we may have to add the optimization to remember validated chunks in order to get at least some of the performance back.

RyuzakiKK: Ah I see. Well this could really negatively impact the performance.

I'm thinking out loud, what would happen if we added Goroutine in `Validate()` too? We could split the segment in X parts and give each one of them to different Goroutine.

folbricht: That is certainly an option. If the candidate has multiple chunks, each of them could be validated concurrently with up to N goroutines. Would have to use the same filehandle as opening new ones per chunk would be too expensive. I suspect this will get back some of the performance, but perhaps not all of it. Combined with a way to remember which chunks have been validated it may get close.

A more efficient and more complex way to do this would be to introduce the concept of a "plan" for how to assemble the file. Could plan it out fully without validation, then run a validation step on the whole plan (concurrently), if a bad seed is found we throw away the plan (and mark the seed bad), and re-run the plan. That what I wanted to do to deal with this originally but never got around to.

folbricht: I can help with that part if you like. And when combining that with remembering what chunks are validated in a seed, this would make the whole process faster than before

RyuzakiKK: Thanks, I guess that your suggestion to use a "plan" is clear enough. I should be able to propose a first implementation "soon", and then we can iterate/improve from that.

RyuzakiKK: I added the concept of the "plan" in my latest revision. I run a couple of quick benchmarks and the difference between this patch and the current master seems to be slim (master ~6m38s and with this patch ~6m55s).

I didn't include the part about remembering which chunks were already validated to avoid complicating too much this PR. But I think that this is definitely something that we should do.

sequencer.go
```diff
@@ -18,18 +26,36 @@ func NewSeedSequencer(idx Index, src ...Seed) *SeedSequencer {
 	}
 }
 
+// Plan returns an array of SeedSegmentCandidate, representing an ordered list of
+// segments that can be used to re-assemble the requested file
+func (r *SeedSequencer) Plan() (plan []SeedSegmentCandidate) {
+	for {
+		seed, segment, source, done := r.Next()
+		plan = append(plan, SeedSegmentCandidate{seed, source, segment})
```
folbricht, line 34: If you look at the previous code, I believe the append should happen after "done" is checked. Looks like when done == true then there is no valid seed in the response (best to verify that). Should be able to make a test where the seed matches the target file and basically the whole file is restored from one segment first, then in the 2nd step "done" should be true but no more seed to use.

fileseed.go
```diff
@@ -22,24 +23,26 @@ func NewIndexSeed(dstFile string, srcFile string, index Index) (*FileSeed, error
...
 }
 
-// LongestMatchWith returns the longest sequence of of chunks anywhere in Source
-// that match b starting at b[0]. If there is no match, it returns nil
+// LongestMatchWith returns the longest sequence of chunks anywhere in Source
+// that match `chunks` starting at chunks[0]. If there is no match, it returns a
+// length of zero and a nil SeedSegment.
 func (s *FileSeed) LongestMatchWith(chunks []IndexChunk) (int, SeedSegment) {
-	if len(chunks) == 0 || len(s.index.Chunks) == 0 {
+	if len(chunks) == 0 || len(s.index.Chunks) == 0 || s.isInvalid {
```
folbricht, line 38: Technically, since you're checking the value of `isInvalid` here, you wouldn't actually need the `IsInvalid()` method in the interface anymore. Currently, `IsInvalid()` is called just before this function. Since that check is redundant. Would let you remove that method from selfseed and nullseed as well.

assemble.go
```diff
@@ -162,19 +259,32 @@ func AssembleFile(ctx context.Context, name string, idx Index, s Store, seeds []
...
+			seq.Rewind()
+			plan = seq.Plan()
+			continue
+		}
+		// Found a valid plan
+		break
+	}
+
+loop:
+	for _, candidates := range plan {
```
folbricht, line 283: Wrt naming, it's not really a "candidate" at this point anymore. Perhaps "segment" would be better?

assemble.go
```diff
@@ -3,11 +3,91 @@ package desync
...
 
+type AssembleOptions struct {
+	N                 int
+	InvalidSeedAction InvalidSeedAction
+}
+
+// validatePlan validates a proposed plan by checking if all the chosen chunks
+// are correctly provided from the seeds. In case a seed has invalid chunks, the
+// entire seed is marked as invalid and an error is returned.
+func validatePlan(ctx context.Context, plan []SeedSegmentCandidate) (err error) {
```
folbricht, line 29: I think this is where you're losing that little bit of performance, and there's significant room for simplification in this function.
1. In seedsequencer.go, you could introduce `type Plan []SeedSegmentCandidate`
2. Then move this function to seedsequencer.go and make it a method of the Plan like so, `func (p Plan) Validate(ctx context.Context, n int)` - you'll need the `n` here for the level of concurrency. You can pass in `N` from the options
3. On the SeedSegment interface, expose the Validate method (it already exists private on the fileseed)
4. Once you have a Validate() method on every segment, you can just call that and you don't have to work with a file descriptor here at all.
5. Open more than one goroutine (loop over `n`) to actually validate the plan concurrently. The `g.Go()` should be in a loop.

Once you have that, I bet the perf difference will disappear completely, and this function will be quite simple and short. If my explanation above doesn't make sense, let me know and I can make a PR for your PR.

assemble.go
```diff
@@ -119,6 +200,22 @@ func AssembleFile(ctx context.Context, name string, idx Index, s Store, seeds []
...
+					if err != nil {
+						return err
+					}
+					stats.addBytesCopied(copied)
+					stats.addBytesCloned(cloned)
+					// Record this segment's been written in the self-seed to make it
+					// available going forward
+					ss.add(job.segment)
+					continue
+				}
```
folbricht, line 217: Hmm, so this won't work quite right since `c` is only the first chunk in a segment to be copied. It'll likely produce invalid file since if there's a segment of 2 chunks to be copied, and the first of those is in the file already, then this will only copy the first chunk and the 2nd one will be skipped, creating a broken output file.

Basically, the introduction of the "plan" has made the self-seed very much useless since it's impossible to use it in the plan. It only gets populated at runtime and it's not currently possible to predict when the selfseed is written, so it can't be part of the plan.

Perhaps the selfseed needs to be retired completely. I wouldn't expect that to have much of a perf impact, especially when a local cache is used.

folbricht: You could actually save the selfseed if `getChunk()` takes a whole segment (not just the first chunk) and then returns an equal segment from itself if one exists of the same length with the same chunks.

sequencer.go
RyuzakiKK: In `assemble.go` there was:
```
		chunks, from, done := seq.Next()
		select {
		case <-ctx.Done():
			break loop
		case in <- Job{chunks, from}:
		}
		if done {
			break
		}
```
And `Next()` returns `segment, source, r.current >= len(r.index.Chunks)`, so the last valid segment is expected to have `done` set to true.

With this info I guess that we need to append and then check the `done`  value.
Does it make sense?

fileseed.go
RyuzakiKK: Right, I removed `IsInvalid()`.

assemble.go
RyuzakiKK: > The g.Go() should be in a loop.

Oops, I forgot to create the `g.Go()` in a loop, so mine was not really doing the validation concurrently.

I implemented what you suggested and now it should be better.

However you mentioned that "you don't have to work with a file descriptor here at all", but I'm not sure if with "here" you meant in `assemble.go` or in the `func (p Plan) Validate()`.
Because to validate a particular `fileSeed` we need a filehandle, and if I open them beforehand and share them to the different Goroutines, we can drastically reduce the amount of `Open()` and `Close()` that we need.

Please let me know if you had something else in mind.

assemble.go
RyuzakiKK: I changed this to `getSegment()` and basically implemented what you suggested.

RyuzakiKK: > Hmm, so this won't work quite right since c is only the first chunk in a segment to be copied. It'll likely produce invalid file since if there's a segment of 2 chunks to be copied, and the first of those is in the file already, then this will only copy the first chunk and the 2nd one will be skipped, creating a broken output file.

I was thinking that this was not failing because when we create a Plan, a segment has more than one chunk only if we are getting it from a seed (and selfseed is not in that list). Like when we get a single chunk from the store.

RyuzakiKK: Probably we should do an assert on the number of chunks that we have? Like if we don't have a seed segment we assert that we just have a single chunk.

If that's not the case we would write a single chunk when reading it from the output file (the `isBlank` portion) and also when we get a it from the store. And if the segment had more than one chunk, the others will be lost.

sequencer.go
folbricht: Oh, yes, you're correct. I got that mixed up with the .Done() from the context. So first append, then check `done` is right

assemble.go
folbricht: So you can actually simplify the selfseed implementation to operate on single chunks, since it's never going to be part of a plan anyway, and won't need the "longestMatch" functionality if I'm reading this right.

folbricht (commented):
Looking pretty good. Did you check performance? I suspect it's about equivalent to before

assemble.go
folbricht: Nope, you got it there, re-using the filehandle for concurrent validation. Looking good.

cmd/desync/extract_test.go
RyuzakiKK: I just noticed that you were commenting on `extract_test`. When you talked about comparing the final hash, were you referring only to the tests cases?

assemble.go
RyuzakiKK: Right. I reverted to the previous `getChunk()` and added a comment to explain this assumption and also a check with `panic()` if that's not the case. If we trip there, there is a programming error.

RyuzakiKK:
> Looking pretty good. Did you check performance? I suspect it's about equivalent to before

I ran a couple of times an extraction and this patch was in average 6 seconds slower (4m30s vs 4m24s) with a cold cache, and about the same speed with a hot cache.
In my test I used a partition with a size of 5gb and the plan was generated in ~7 seconds (with the cache dropped `echo 3 | sudo tee /proc/sys/vm/drop_caches`). With a hot cache it takes less than a second.

Apparently the downside is that now the progress bar doesn't report any progress for the first few seconds, while it generates the plan.
Unfortunately I don't think it's possible to do a realistic estimation regarding how much this step is gonna take compared to the whole update.

IMHO that's fine, but if you'd like to show some sort of progress we can figure something out.

cmd/desync/extract_test.go
folbricht: Yes, but that can also wait for another PR. I just think given how easy it is to get the seed-algorithms wrong, it'd be good to check that the whole file is assembled correctly in tests. It'd be a little tricky to come up with good test seeds though, since it should cover file-seeds, self-seeds, null-seed etc in specific combinations to exercise all paths properly. On the other hand, having a predicatable seed plan now (it was non-deterministic before) should help.

assemble.go
```diff
@@ -118,7 +135,30 @@ func AssembleFile(ctx context.Context, name string, idx Index, s Store, seeds []
...
+				if segment := ss.getChunk(c.ID); segment != nil {
+					copied, cloned, err := segment.WriteInto(f, c.Start, c.Size, blocksize, isBlank)
+					if err != nil {
+						return err
+					}
+					stats.addBytesCopied(copied)
+					stats.addBytesCloned(cloned)
+					// Record this segment's been written in the self-seed to make it
+					// available going forward
+					ss.add(job.segment)
```
folbricht, line 158: Interesting, this wouldn't even be necessary anymore, since we just confirmed we have the chunk in the self-seed already and it's all operating on a single chunk. It won't hurt of course, would only be useful if there was a whole segment we wanted to take from the selfseed. But it might be good to keep it and avoid any surprises when the logic is changed to support whole sections of self-seed. Should we update the comment to say it's not actually needed?

folbricht (commented):
Great work. Please take a look at my last comment and then I think this is good to merge in.

As for the progressbar, I agree that it's not critical to have right now. It'd be quite simple to wire up and introduce a whole step in it. If you look at `desync make` for example, you can see there's multiple stages in the progressbar. So you could make "Planning" and "Validating" stages themselves in the progressbar and calculate it based on the # of chunks to get through. Planning would be very quick, I suspect the 3s you saw was spent in the validation step.

assemble.go
RyuzakiKK: Good point. I removed the `ss.add()` and wrote a comment to explain why it is not needed.
In this way, if in the future we add support for whole segments, it will be clear enough that the assumption of a single chunk would not be valid anymore.

assemble.go
```diff
@@ -118,7 +135,30 @@ func AssembleFile(ctx context.Context, name string, idx Index, s Store, seeds []
...
+					copied, cloned, err := segment.WriteInto(f, c.Start, c.Size, blocksize, isBlank)
+					if err != nil {
+						return err
+					}
+					stats.addBytesCopied(copied)
+					stats.addBytesCloned(cloned)
+					// This is just a single chunk, and we already confirmed it to be
+					// present in the self-seed. For this reason there is no need to
+					// record this segment again as being written in the self-seed.
+					continue
```
folbricht, line 159: Sorry, I may have been wrong about not needing to "add" to the selfseed here. Take a look at https://github.com/folbricht/desync/blob/master/selfseed.go#L45-L65 So basically, the selfseed previously supported matching on multi-chunk segments, similar to the other seeds. Now, to achieve that it kept a pointer into the index describing up to which point the selfseed is fully populated. Only the segments below this pointer could be used for seeding. So now if we don't "add" all the chunks to it, the pointer won't advance past a certain stage and the seed will be less useful. We could:
1) Keep that code in selfseed and put the "add" back here
2) Or just remove that logic from the self-seed and have a different way of remembering which chunks are available in the selfseed.

I suspect that 1) may be the easiest option at this point, but this could be revisited in another PR. Thoughts?
(Sorry for not realizing that issue earlier)

RyuzakiKK: Ops, you suggestion made a lot of sense and I implemented it without double checking the side effects. Sorry about that.

I implemented 1), adding back the `ss.add()` and explaining why we need that.
I prefer not to add too much changes in a single commit. With a followup MR we can always refactor how the self-seed is generated and maybe optimize it a little more by avoiding the need to add chunks that are already in the self-seed.

folbricht (approved):

folbricht:
Thank you for working through all of this

PR #205, opened 2021-12-07
Add a goreleaser configuration and workflow

### LINKED ISSUE #204: enhancement: binary releases
Would you be open to a PR which adds goreleaser (or similar) support for cross-building binaries and hosting them on Github Releases?

I find I'm quite often pulling desync into containers or build agents where I don't have a full go setup, and if I'm going to have to host binaries of desync for my own purposes, I thought it might be useful to have some open-source hosted binaries available for everyone.

This would also allow us to make a brew package available fairly easily (as a future enhancement), if that was wanted

dominics:
Adds a .goreleaser.yml and a Github Actions workflow to do a release when a new tag is pushed. Fixes #204

The config was generated with `goreleaser init`, then modified to:
 - turn on GO111MODULE
 - point at cmd/desync as main
 - ignore old arm64 versions on go 1.15 (see below)
 
This configuration is using the same Go version as the verify workflow (1.15), so windows/arm64 and darwin/arm64 platforms are ignored until we're using Go 1.16/1.17 (so, no Apple Silicon build yet). This [ignore happens anyway in the latest versions of goreleaser](https://goreleaser.com/deprecations/#builds-for-darwinarm64), but wasn't happening on the slightly older version of goreleaser included in the action workflow (so, we'd [fail like this](https://github.com/dominics/desync/runs/4450617547?check_suite_focus=true#step:4:74))

This configuration runs [a release workflow on tag](https://github.com/dominics/desync/runs/4450669985?check_suite_focus=true#step:4:71), and produces [releases that look something like this](https://github.com/dominics/desync/releases/tag/v0.9.3-rc.5) (but with a slightly nicer changelog once this squash-merges)

image


folbricht (approved):
Thank you. I think we should update go.mod to require Go 1.16, especially if that allows producing binaries for more platforms automatically. Would you like to make a PR for that?

PR #206, opened 2021-12-21
Speed up VerifyIndex

RyuzakiKK:
- Add some tests for VerifyIndex
- Reduce code duplication while verifying an index
    We can reuse the validate() function of FileSeedSegment and avoid
duplicating the code here.
- Feed multiple chunks to the workers while verifying an index
    If we send a single chunk to every Goroutine, the workers will complete
their task immediately and then will wait for a new chunk to validate.

    If instead we group together multiple chunks we can improve the overall
performance.

    For example, in my test machine, the required time to verify an index
that pointed to a 5gb image took 30 seconds, and with this patch it
dropped to 24 seconds.

### DIFF
--- a/cmd/desync/verifyindex_test.go
+++ b/cmd/desync/verifyindex_test.go
@@ -1,0 +1,34 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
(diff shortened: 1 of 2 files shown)


folbricht (approved):
Great re-use of the seed validation logic.

PR #207, opened 2021-12-23
Get the correct size for block devices

RyuzakiKK:
The function `Size()` of `FileInfo` returns the length in bytes for
regular files. However if we use a block device, `Size()` will return a
length of zero.

To chunk a file, we use the size to split the expected work in even
parts.

For this reason, when we wanted to generate an index from a block
device, due to the incorrect reported zero size, the whole task was
carried out by a single Goroutine.

With this commit we use the ioctl `BLKGETSIZE64` to get the correct size
for block devices.

In a test environment the `make` operation against a block device took 2
minutes and 20 seconds to complete with the current master branch and
only 24 seconds with this patch.

### DIFF
--- a/ioctl_linux.go
+++ b/ioctl_linux.go
@@ -14,6 +14,9 @@ import (
 	"github.com/pkg/errors"
 )
 
+// BLKGETSIZE64 ioctl
+const blkGetSize64 = 0x80081272
+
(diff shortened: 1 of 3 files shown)


folbricht (approved):
Nice one!

PR #209, opened 2022-01-06
Support glob patterns in the configuration file 

RyuzakiKK:
- Bail out if multiple config entries match a single location
    If we have more than one entry in the configuration file that match the
location string that we are currently using (e.g. one with a trailing
slash and another without), we will end up choosing one configuration at
random.

    When this situation arise, we should instead abort the execution to
avoid unexpected results.

- Support glob patterns in the configuration file
    This makes the configuration file more flexible.

The first commit has the potential to break backwards compatibility. Right now a configuration file that had both `/foo/bar` and `/foo/bar/` was acceptable, and at runtime Desync was choosing one of them in an undefined order.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,8 +241,8 @@ Available configuration values:
 
 - `http-timeout` *DEPRECATED, see `store-options.<Location>.timeout`* - HTTP request timeout used in HTTP stores (not S3) in nanoseconds
 - `http-error-retry` *DEPRECATED, see `store-options.<Location>.error-retry` - Number of times to retry failed chunk requests from HTTP stores
-- `s3-credentials` - Defines credentials for use with S3 stores. Especially useful if more than one S3 store is used. The key in the config needs to be the URL scheme and host used for the store, excluding the path, but including the port number if used in the store URL. It is also possible to use a [standard aws credentials file](https://docs.aws.amazon.com/cli/latest/userguide/cli-config-files.html) in order to store s3 credentials.
-- `store-options` - Allows customization of chunk and index stores, for example comression settings, timeouts, retry behavior and keys. Not all options are applicable to every store, some of these like `timeout` are ignored for local stores. Some of these options, such as the client certificates are overwritten with any values set in the command line. Note that the store location used in the command line needs to match the key under `store-options` exactly for these options to be used. Watch out for trailing `/` in URLs.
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -241,8 +241,8 @@ Available configuration values:
...
-- `store-options` - Allows customization of chunk and index stores, for example comression settings, timeouts, retry behavior and keys. Not all options are applicable to every store, some of these like `timeout` are ignored for local stores. Some of these options, such as the client certificates are overwritten with any values set in the command line. Note that the store location used in the command line needs to match the key under `store-options` exactly for these options to be used. Watch out for trailing `/` in URLs.
+- `s3-credentials` - Defines credentials for use with S3 stores. Especially useful if more than one S3 store is used. The key in the config needs to be the URL scheme and host used for the store, excluding the path, but including the port number if used in the store URL. The key can also contain glob patterns, and the available wildcards are `*`, `?` and `[…]`. Please refer to the [filepath.Match documentation] for additional information. It is also possible to use a [standard aws credentials file](https://docs.aws.amazon.com/cli/latest/userguide/cli-config-files.html) in order to store s3 credentials.
```
folbricht, line 244: you can use `[filepath.Match](https://pkg.go.dev/path/filepath#Match)` to create a direct link

cmd/desync/config.go
```diff
@@ -70,14 +70,21 @@ func (c Config) GetS3CredentialsFor(u *url.URL) (*credentials.Credentials, strin
...
+// an error will be returned if the location string matches multiple entries in the
+// config file.
+func (c Config) GetStoreOptionsFor(location string) (options desync.StoreOptions, err error) {
+	found := false
+	options = desync.StoreOptions{}
 	for k, v := range c.StoreOptions {
 		if locationMatch(k, location) {
-			return v
+			if found {
+				return options, fmt.Errorf("multiple configuration entries match the location '%s'", location)
```
folbricht, line 81:
```suggestion
				return options, fmt.Errorf("multiple configuration entries match the location %q", location)
```

folbricht (commented):
Looking good. Thank you

README.md
RyuzakiKK: Ops, I forgot to put the actual link. Fixed now.

folbricht (approved):

PR #211, opened 2022-01-17
Add new ENV variable to make it easier to parse the progresses

RyuzakiKK:
Introduce a new environment variable `DESYNC_ENABLE_PARSABLE_PROGRESS`
that allows us to print in output the progress without the progress bar.

This will make it easier to parse the current completion percentage, and
estimated remaining time, without flooding the journal, or where the
stderr was redirected, with very frequent updates that contain long
progress bars.

-----

In theory I could have reused the same `DESYNC_PROGRESSBAR_ENABLED`, but I didn't want to change behavior in case other people were already parsing its output. Also an env var called "progressbar enabled" that doesn't actually print a progress bar was probably too awkward.


folbricht (approved):

folbricht:
Thank you

PR #213, opened 2022-02-24
Fix inline comment about the error-retry-base-interval

RyuzakiKK:
As correctly mentioned in the README.md, this option is in nanoseconds
and the default value is zero.

---

This is a minor fix. However I already tripped over it, so I guess that it's worth to correct this comment.


folbricht (approved):
Thank you

PR #214, opened 2022-02-25
Validate the written chunks when using a seed

RyuzakiKK:
The seed values are validated at the beginning when we create the
extraction "plan".

However a seed might point to a location where is possible to read and
write data. In that case, even if at the beginning we validated the seed
data, the assumption that when we use `WriteInto()` the seed data is
still valid might not be true anymore.

For this reason we should also re-validate the written data to ensure
that what ended up in the destination is exactly what we were expecting.

### DIFF
--- a/assemble.go
+++ b/assemble.go
@@ -128,6 +128,22 @@ func AssembleFile(ctx context.Context, name string, idx Index, s Store, seeds []
 					if err != nil {
 						return err
 					}
+
+					// Validate that the written chunks are exactly what we were expecting.
+					// Because the seed might point to a RW location, if the data changed
+					// while we were extracting an index, we might end up writing to the
+					// destination some unexpected values.
+					for _, c := range job.segment.chunks() {
+						b := make([]byte, c.Size)
+						if _, err := f.ReadAt(b, int64(c.Start)); err != nil {
+							return err
+						}
+						sum := Digest.Sum(b)
+						if sum != c.ID {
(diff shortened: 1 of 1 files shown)


RyuzakiKK:
This should have a negligible impact on performance. However if you prefer I can move this additional check under a launch option.

assemble.go
```diff
@@ -128,6 +128,22 @@ func AssembleFile(ctx context.Context, name string, idx Index, s Store, seeds []
...
+					// while we were extracting an index, we might end up writing to the
+					// destination some unexpected values.
+					for _, c := range job.segment.chunks() {
+						b := make([]byte, c.Size)
+						if _, err := f.ReadAt(b, int64(c.Start)); err != nil {
+							return err
+						}
+						sum := Digest.Sum(b)
+						if sum != c.ID {
+							return fmt.Errorf("written data in %s doesn't match its expected value", name)
```
folbricht, line 143: Since at this point we can be pretty sure that a seed was modified during processing, we could be more clear in the message to help anyone that sees this message and debugging it. How about this?

`written data in %s doesn't match its expected hash value, seed may have changed during processing`

folbricht:
I agree that it shouldn't add to much overhead since the data was just written and it should be read back from buffered memory.

assemble.go
RyuzakiKK: Right, I changed the error message with your suggested string.

folbricht (approved):

PR #216, opened 2022-03-25
Add option to regenerate invalid seed indexes

RyuzakiKK:
Sometimes we don't have the guarantee that a particular seed index is
always valid.
For example if a seed index is calculated and used in two different
occasions, it might happen that in the meantime some of the seed files
got corrupted or that they simply changed (if the seed is a writable
location).

One workaroud is to manually launch `verify-index` and then regenerate
the index if that check fails. However this involves additional steps
and is potentially slower because we need to verify the entire seed
before running the `extract` command.

Instead with the new option "--regenerate-invalid-seeds", Desync will
automatically regenerate the seed index if the validation step fails.


fileseed.go
```diff
@@ -63,12 +64,34 @@ func (s *FileSeed) LongestMatchWith(chunks []IndexChunk) (int, SeedSegment) {
...
 }
 
+func (s *FileSeed) RegenerateIndex(ctx context.Context, n int) error {
+	index, _, err := IndexFromFile(ctx, s.srcFile, n, s.index.Index.ChunkSizeMin, s.index.Index.ChunkSizeAvg,
+		s.index.Index.ChunkSizeMax, nil)
+	if err != nil {
+		return err
+	}
+
+	s.index = index
```
folbricht, line 74: Don't you also need to call `SetInvalid(false)` after resetting it?

sequencer.go
```diff
@@ -76,6 +76,19 @@ func (r *SeedSequencer) Rewind() {
...
 }
 
+// RegenerateInvalidSeeds regenerates the index to match the unexpected seed content
+func (r *SeedSequencer) RegenerateInvalidSeeds(ctx context.Context, n int) error {
+	for _, s := range r.seeds {
+		if s.IsInvalid() {
+			if err := s.RegenerateIndex(ctx, n); err != nil {
+				return err
+			}
+			s.SetInvalid(false)
```
folbricht, line 86: oh, I see it's being done here. Either works but it might be better to do it in the seed since that's where the regenerate happens

assemble.go
```diff
@@ -227,9 +230,16 @@ func AssembleFile(ctx context.Context, name string, idx Index, s Store, seeds []
 			// This plan has at least one invalid seed
 			if options.InvalidSeedAction == InvalidSeedActionBailOut {
 				return stats, err
+			} else if options.InvalidSeedAction == InvalidSeedActionRegenerate {
```
folbricht, line 233: Almost time for a `switch options.InvalidSeedAction {` here

cmd/desync/extract.go
```diff
@@ -52,6 +53,7 @@ the index from STDIN.`,
 	flags.StringSliceVar(&opt.seeds, "seed", nil, "seed indexes")
 	flags.StringSliceVar(&opt.seedDirs, "seed-dir", nil, "directory with seed index files")
 	flags.BoolVar(&opt.skipInvalidSeeds, "skip-invalid-seeds", false, "Skip seeds with invalid chunks")
+	flags.BoolVar(&opt.regenerateInvalidSeeds, "regenerate-invalid-seeds", false, "Regenerate seed indexes with invalid chunks")
```
folbricht, line 56: It might be good to add something to the command description itself. The reason is that "regenerate" could be understood in two ways.
1) Keep the data and build an index from it in memory, neither data nor index are changed on disk (worth mentioning too). This is what is implemented
2) The index is used to restore the seed data file - This is not what's happening but it'd be good to make sure it's not understood that way

folbricht (approved):

sequencer.go
RyuzakiKK: Yes, it's probably better to do that in `fileseed`. I moved it.

assemble.go
RyuzakiKK: Done, with a default that panics to ensure that we are actually considering all the possibilities.

cmd/desync/extract.go
RyuzakiKK: Right, I explained it in the extract description.

RyuzakiKK:
I also added the commit [Add the correct seed in the extract test](https://github.com/folbricht/desync/pull/216/commits/741c26c2f2bc809217f0b7772f6ce3c81e2043b9), because I noticed that from time to time that test was failing.

It can be moved to its own PR if you prefer.

folbricht (approved):
Thank you

PR #217, opened 2022-03-30
Create a progress bar for the plan validation

RyuzakiKK:
Validating a plan could take a non trivial amount of time, depending on
the size of a plan and by the I/O speed.

With this commit we add a progress bar that shows the status of the
validation and prefixes the assembling progress bar with "Assembling ".

---

I added just the validation and not the actual "planning" because, from my testings, the planning is usually done nearly instantaneously and the progress percentage would just go from 0% to 100%.

### DIFF
--- a/progressbar.go
+++ b/progressbar.go
@@ -15,7 +15,7 @@ func NewProgressBar(prefix string) ProgressBar {
 	if !terminal.IsTerminal(int(os.Stderr.Fd())) &&
 		os.Getenv("DESYNC_PROGRESSBAR_ENABLED") == "" &&
 		os.Getenv("DESYNC_ENABLE_PARSABLE_PROGRESS") == "" {
-		return nil
+		return NullProgressBar{}
 	}
 	bar := pb.New(0).Prefix(prefix)
 	bar.ShowCounters = false
--- a/progressbar.go
+++ b/progressbar.go
@@ -1,18 +1,17 @@
-package main
+package desync
 
 import (
 	"fmt"
 	"os"
 	"time"
 
-	"github.com/folbricht/desync"
 	"golang.org/x/crypto/ssh/terminal"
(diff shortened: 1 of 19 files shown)


progressbar.go
```diff
@@ -31,30 +30,30 @@ func NewProgressBar(prefix string) desync.ProgressBar {
 		bar.Callback = func(s string) { fmt.Fprintln(os.Stderr, s) }
 		bar.Output = nil
 	}
-	return ProgressBar{bar}
+	return ProgressBarDesync{bar}
 }
 
-// ProgressBar wraps https://github.com/cheggaaa/pb and implements desync.ProgressBar
-type ProgressBar struct {
+// ProgressBarDesync wraps https://github.com/cheggaaa/pb and implements desync.ProgressBar
```
folbricht, line 36: Wrt naming, would `DefaultProgressBar` work as well? Not ideal to suffix it with `Desync` tbh.

folbricht (commented):
Looks good, thank you. If you want there's an opportunity to improve my old code here. If `NewProgressBar()` returned something like `NullProgressBar` that implements the `ProgressBar` interface instead of nil in the non-terminal case, you could get rid of all the nil checks for it thoughout the code (the `if pb != nil` bit)

RyuzakiKK:
Sure, I should be able to do that tomorrow.

RyuzakiKK:
With the latest revision I added the suggested `NullProgressBar`, and the progressbar for when we chunk an invalid seed.

folbricht (approved):
Thank you

folbricht:
Hmm, looks like the MacOS build breaks due to a lib issue, likely need some built-tags.

folbricht:
Nope, unrelated build issue in Go 1.18. Fixed in https://github.com/folbricht/desync/pull/218

PR #219, opened 2022-04-11
Write the progress bar attempt number

RyuzakiKK:
Currently we have three major steps while extracting a bundle:
validation, seed chunking and assembling.

If one or more seeds are invalid, these steps might be repeated until
Desync is able to make a plan that is correct.

To make it easier to parse and understand what is going on, we prepend
the attempt number to the progress bar description.

In this way, for example, we will have an output like the following:
Attempt 1: Validating    0.00%
Attempt 1: Validating    0.45% 0s
Attempt 2: Chunking Seed 1    0.00%
Attempt 2: Chunking Seed 1    0.27% 00m35s
Attempt 2: Chunking Seed 1  100.00% 12s
Attempt 2: Validating    0.00%
Attempt 2: Validating   30.43% 00m01s
Attempt 2: Validating  100.00% 1s
Attempt 2: Assembling    0.00%
[...]

### DIFF
--- a/assemble.go
+++ b/assemble.go
@@ -39,14 +39,13 @@ func AssembleFile(ctx context.Context, name string, idx Index, s Store, seeds []
 		source  SeedSegment
 	}
 	var (
+		attempt     = 1
 		in          = make(chan Job)
 		isBlank     bool
 		isBlkDevice bool
 		pb          ProgressBar
 	)
 	g, ctx := errgroup.WithContext(ctx)
-
-	pb = NewProgressBar("Assembling ")
 
 	// Initialize stats to be gathered during extraction
 	stats := &ExtractStats{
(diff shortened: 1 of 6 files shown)


evelikov:
One small comment - other than that, I this looks great.

I think the example is a little off - validating is always the first step, isn't it?

```
Attempt 1: Validating 0.00%
Attempt 1: Validating 0.45% 0s
Attempt 1: Chunking Seed 1 0.00% // this and next couple of lines were "2"
Attempt 1: Chunking Seed 1 0.27% 00m35s
Attempt 1: Chunking Seed 1 100.00% 12s
Attempt 2: Validating 0.00%
...
```

RyuzakiKK:
No, that reflects the current implementation where I increase the attempt number after a failed validation.

The reasoning behind it is that if a validation fails, I considered chunking a seed (i.e. fixing it) already the attempt 2, because our first attempt already failed.

Did you prefer having the chunking in the same attempt as the validation, to keep the validations always as the first step?

evelikov:
As an outsider (aka someone who doesn't know the desync internals), it makes sense to have each attempt encapsulated in the validate/chunk/assemble order. For example:
```
Attempt 1: Validate // oops need chunking
Attempt 1: Chunk
Attempt 2: Validate
Attempt 2: Assemble // oops writing has failed due to reasons
Attempt 3: Validate // disk has changed, need chunking
Attempt 3: Chunking
Attempt 4: Validate
Attempt 4: Assemble // everything was successful
```

At a glance the above matches the actual execution order within desync as well. So it maps nicely AFAICT.

folbricht (approved):

folbricht:
Thank you

PR #220, opened 2022-04-15
Do not immediately abort the extraction if a seed chunk is invalid 

RyuzakiKK:
- Factor out writeChunk to a separate function
- Do not immediately abort the extraction if a seed chunk is invalid

    If the seed points to a RW location it may happen that some files change
while we are in the middle of an extraction.

    When using the "InvalidSeedActionRegenerate" option, we can try harder
and attempt to take the invalid chunks from the self seed or the store.
If both of those fail too, then we abort the entire operation.

- Limit the maximum number of chunks is a single seed sequence

    If a seed is nearly completely equal to the output, we may end up with
just a few SeedSequencer that are very long and others that are just
one/a few chunks long.

    Because each SeedSequencer is handled by a goroutine (from a pool), we
may reach a situation where the majority of the goroutines finish their
operations and are just waiting for the longer SeedSequencer jobs to
end.

    By limiting the maximum amount of chunks in the SeedSequencer, we will
have jobs that are more balanced in term of amount of work.

- When taking a chunk from the self seed, immediately return

    If we were able to take a chunk from the self seed, there is no need to
    continue looking into the existing file and the store.

### DIFF
--- a/assemble.go
+++ b/assemble.go
@@ -25,6 +25,57 @@ type AssembleOptions struct {
 	InvalidSeedAction InvalidSeedAction
 }
 
+// writeChunk tries to write a chunk by looking at the self seed, if it is already existing in the
+// destination file or by taking it from the store
+func writeChunk(c IndexChunk, ss *selfSeed, f *os.File, blocksize uint64, s Store, stats *ExtractStats, isBlank bool) error {
+	// If we already took this chunk from the store we can reuse it by looking
+	// into the selfSeed.
+	if segment := ss.getChunk(c.ID); segment != nil {
+		copied, cloned, err := segment.WriteInto(f, c.Start, c.Size, blocksize, isBlank)
+		if err != nil {
+			return err
+		}
+		stats.addBytesCopied(copied)
+		stats.addBytesCloned(cloned)
+	}
+
+	// If we operate on an existing file there's a good chance we already
+	// have the data written for this chunk. Let's read it from disk and
+	// compare to what is expected.
+	if !isBlank {
+		b := make([]byte, c.Size)
+		if _, err := f.ReadAt(b, int64(c.Start)); err != nil {
+			return err
+		}
+		sum := Digest.Sum(b)
+		if sum == c.ID {
+			// Record we kept this chunk in the file (when using in-place extract)
+			stats.incChunksInPlace()
+			return nil
+		}
+	}
+	// Record this chunk having been pulled from the store
+	stats.incChunksFromStore()
+	// Pull the (compressed) chunk from the store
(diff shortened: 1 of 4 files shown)


RyuzakiKK:
I manually tested the situation where a seed changes while `desync extract` is in progress.

However I'd like to add some automated tests too. I'll probably look into it in the days.

assemble.go
```diff
@@ -137,7 +189,15 @@ func AssembleFile(ctx context.Context, name string, idx Index, s Store, seeds []
 						}
 						sum := Digest.Sum(b)
 						if sum != c.ID {
-							return fmt.Errorf("written data in %s doesn't match its expected hash value, seed may have changed during processing", name)
+							if options.InvalidSeedAction == InvalidSeedActionRegenerate {
+								// Try harder before giving up and aborting
+								Log.WithField("ID", c.ID).Info("The seed may have changed during processing, trying to take the chunk from the self seed or the store")
+								if err := writeChunk(c, ss, f, blocksize, s, stats, isBlank); err != nil {
+									return err
+								}
```
RyuzakiKK, line 197: This could be further optimized, for example by remembering which chunks were invalid and avoiding to rewriting them.
I didn't do it for now because it would increase the complexity of this PR and I wanted to start with something simpler and maybe add on top of it later on.

folbricht: Given that this only happens in (hopefully rare) situations it's perfectly fine to not worry about efficiency. No need to worry about efficiency in the failure-case.

sequencer.go
```diff
@@ -55,15 +55,23 @@ func (r *SeedSequencer) Next() (seed Seed, segment IndexSegment, source SeedSegm
 	var (
 		max     uint64
 		advance = 1
+		// Limit the maximum number of chunks, in a single sequence, to avoid
+		// having jobs that are too unbalanced
+		limit = 100
```
folbricht, line 60: There are situations where this will degrade performance. Specifically when running on a filesystem that supports reflinks. There, this limit would make it slower as well as take up more disk space. I wonder if it'd be better to remove the limit parameter from the interface and instead put a limit into the nullseed, as well as the fileseed if the FS does not support cloning. The fileseed already has the `canReflink` flag which could be used to set an infinite level.
Removing the limit from the interface would be cleaner as well.

assemble.go
RyuzakiKK: Right, I totally agree.

sequencer.go
RyuzakiKK: Oh I didn't realize that.

I implemented your suggestion and now the limit is applied only if a seed can't be reflinked.

folbricht (approved):

PR #221, opened 2022-04-18
Test a seed that changes while an extraction is in progress

RyuzakiKK:
This adds an automated tests for #220.

I added an environment variable in `sequencer` because that seemed the cleaner way to mock  the `Validate()` function.
Please let me know if instead there is a better way to do that.

### DIFF
--- a/sequencer.go
+++ b/sequencer.go
@@ -26,6 +26,8 @@ type SeedSegmentCandidate struct {
 
 type Plan []SeedSegmentCandidate
 
+var MockValidate = false
+
 // NewSeedSequencer initializes a new sequencer from a number of seeds.
 func NewSeedSequencer(idx Index, src ...Seed) *SeedSequencer {
 	return &SeedSequencer{
@@ -108,6 +110,10 @@ func (p Plan) Validate(ctx context.Context, n int, pb ProgressBar) (err error) {
 		in      = make(chan Job)
 		fileMap = make(map[string]*os.File)
 	)
+	if MockValidate {
+		// This is used in the automated tests to mock a plan that is valid
+		return nil
+	}
 	length := 0
 	for _, s := range p {
 		if !s.isFileSeed() {
--- a/assemble_test.go
+++ b/assemble_test.go
@@ -5,9 +5,11 @@ import (
 	"context"
 	"crypto/md5"
 	"crypto/rand"
+	"github.com/stretchr/testify/require"
 	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"testing"
 )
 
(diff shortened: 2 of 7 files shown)


cmd/desync/extract_test.go
```diff
@@ -154,3 +154,44 @@ func TestExtractWithInvalidSeeds(t *testing.T) {
...
+		output string
+	}{
+		{"extract with seed that becomes invalid while extracting",
+			[]string{"--store", "testdata/blob2.store", "--seed", "testdata/blob2_corrupted.caibx", "--regenerate-invalid-seeds"}, "testdata/blob2.caibx", out},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			// Assume the seed to be okay while validating the plan
+			// When targeting at least Go 1.17, we can replace the following with "t.Setenv()"
+			err := os.Setenv("DESYNC_MOCK_VALID_PLAN", "1")
+			t.Cleanup(func() { os.Unsetenv("DESYNC_MOCK_VALID_PLAN") })
```
RyuzakiKK, line 177: Go 1.17 introduced scoped env vars, so these two lines could be replaced with `t.Setenv("DESYNC_MOCK_VALID_PLAN", "1")`

Github VMs seems to include Go 1.17 now https://github.com/actions/virtual-environments/issues/5280 but in `go.mod` the minimum Go version is the 1.15.
Probably not worth it to bump the Go version for just this minor improvement?

folbricht: Yes, it's definitely time to bump the min version to 1.17 since 1.18 is out already. Should also update the dependencies at that time.

cmd/desync/extract_test.go
```diff
@@ -154,3 +154,44 @@ func TestExtractWithInvalidSeeds(t *testing.T) {
 		})
 	}
 }
+
+func TestExtractWithNonStaticSeeds(t *testing.T) {
+	outDir, err := ioutil.TempDir("", "")
+	require.NoError(t, err)
+	defer os.RemoveAll(outDir)
+	out := filepath.Join(outDir, "out")
```
folbricht, line 162: This is in line with the rest of the code, but we could start using https://pkg.go.dev/testing#T.TempDir which was introduced in 1.15.

sequencer.go
```diff
@@ -108,6 +108,10 @@ func (p Plan) Validate(ctx context.Context, n int, pb ProgressBar) (err error) {
 		in      = make(chan Job)
 		fileMap = make(map[string]*os.File)
 	)
+	if mock := os.Getenv("DESYNC_MOCK_VALID_PLAN"); mock != "" {
```
folbricht, line 111: Not a fan of injecting testing variables via the environment. Let's think of alternatives:
1) The test could move into the `desync` package rather than `main`. This would provide more control over what is mocked. One could use an interface with a testing-implementation, or a non-exported global variable that the test sets before running
2) Exporting a variable that can be set from `main`. That's not ideal though.
Any other ideas?

RyuzakiKK: Thanks for the suggestions!
I'll research a bit more about the possible solutions and I'll try to improve this PR.

cmd/desync/extract_test.go
RyuzakiKK: Now that I don't use the environment variable anymore, we can probably bump the min version in a separate PR.

cmd/desync/extract_test.go
RyuzakiKK: Yeah that's definitely useful. In the latest revision I started to use it.

sequencer.go
RyuzakiKK: After reading the Go documentation and a few articles online, I ended up using a non-exported global variable to flag the mock validation.

Please let me know if you preferred a different approach instead.

RyuzakiKK:
I guess this PR fell through the cracks. Is it Okay as-is or would you prefer some changes?

folbricht:
Thanks for following up, forgot about this.

PR #223, opened 2022-08-24
feat: add HTTP cookie header support

oliwilkinsonio:
Adds a `HTTPCookie` option to the RemoteHttpStore.

Allows use of desync with a Google Cloud CDN that uses the [Signed Cookies](https://cloud.google.com/cdn/docs/using-signed-cookies) feature.

### DIFF
--- a/remotehttp.go
+++ b/remotehttp.go
@@ -123,11 +123,6 @@ func (r *RemoteHTTPBase) IssueHttpRequest(method string, u *url.URL, getReader G
 	if r.opt.HTTPCookie != "" {
 		req.Header.Set("Cookie", r.opt.HTTPCookie)
 	}
-	// Set from env var if appropriate
-	httpCookie := os.Getenv("DESYNC_HTTP_COOKIE")
-	if len(httpCookie) != 0 {
-		req.Header.Set("Cookie", httpCookie)
-	}
 
 	log.Debug("sending request")
 	resp, err = r.client.Do(req)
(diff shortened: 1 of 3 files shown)


remotehttp.go
```diff
@@ -119,6 +120,14 @@ func (r *RemoteHTTPBase) IssueHttpRequest(method string, u *url.URL, getReader G
 	if r.opt.HTTPAuth != "" {
 		req.Header.Set("Authorization", r.opt.HTTPAuth)
 	}
+	if r.opt.HTTPCookie != "" {
+		req.Header.Set("Cookie", r.opt.HTTPCookie)
+	}
+	// Set from env var if appropriate
+	httpCookie := os.Getenv("DESYNC_HTTP_COOKIE")
```
folbricht, line 127: This basically means the env var takes precedence over the config file. It might be good to swap the logic around.

folbricht (commented):
Could you also add the new option to the README.md? https://github.com/folbricht/desync#configuration

remotehttp.go
oliwilkinsonio: Good shout, will change that.

oliwilkinsonio: In fact, I'll remove it. The previous `DESYNC_HTTP_AUTH` header is only used for the server. I could add it to the server, but not sure how much sense that makes!

folbricht (approved):
Thank you

folbricht:
Oh, there's an unused import which triggers a build failure. Can you remove that?

PR #227, opened 2022-11-30
Fix untar progress percentage

RyuzakiKK:
If we start the progress bar before setting its total size, the progress bar will keep by default a hidden progress percentage. So we will only get the progress animation without the written percentage.

To fix that we should instead set the total before starting the pb.


folbricht (approved):

folbricht:
Thank you

PR #229, opened 2022-12-03
allow store to use http query params in requests

om26er:
The chunk store does not pass http query params when downloading chunks, this means if a URL is protected desync won't be able to download and hence fail. This PR fixes that and I have verified the change actually does work with Cloudfront's presigned URLs


om26er:
@folbricht what do you think about this ?

folbricht:
This basically preserves any query parameters given in the store URL (for HTTP stores and index stores). This seems safe to do. Would you be able to add a section in the README.md describing how to use signed URLs with a simple example?

om26er:
Sure, I will add something during the weekend

Issue #228, opened 2022-12-02
Add a `--seed-and-data` option or something similar


RyuzakiKK:
When using the `--seed foo.caibx` option, Desync expects to find the data in `foo`, next to the provided `.caibx` file.

While this is a valid assumption for files, it requires additional steps when the data is a block device. For example a common suggestion is to create a symlink next to the seed index.

I was thinking that maybe we could have an additional option, e.g. `--seed-and-data`, that takes the seed index and seed data in the same argument. Something like `--seed-and-data /path/to/seed.caibx:/path/to/seed_data`.

Another alternative could be to have ordered command line options?

Before implementing it, @folbricht I'd like to hear you opinion first.

folbricht:
It's quite unfortunate that we need a new parameter for this, but I guess there's no real way to reuse the exiting `--seed` while maintaining 100% backwards compatibility. Would have been nice to do something like `--seed file.caibx@somefile`, or as in your case `:`.

Ordered command lines aren't pretty either. How about reading the seed data from a text file that has one line per seed like
```
file.caibx:data.bin:<perhaps other options in the future>
```
This could even be read from stdin if we wanted to.

RyuzakiKK:
> It's quite unfortunate that we need a new parameter for this, but I guess there's no real way to reuse the exiting --seed while maintaining 100% backwards compatibility. Would have been nice to do something like --seed file.caibx@somefile, or as in your case :.

I was thinking that maybe we could even do:

- If the `--seed` value ends with `.caibx`, it is treated as a seed index path (as Desync does now). Its data will be searched next to it, without the `.caibx` extension.
- If the `--seed` value has `@` (or even `.caibx@`), split the argument at `@`. Then use the first part as the seed index path and the second as the seed data.

Given that `--seed` currently works only with paths that end with `.caibx`, this will not be a breaking change for compatibility.

(and do you prefer `@` or `:`? Probably, as you mentioned, `:` is better to leave open the possibility for additional options in the future)

folbricht:
I like it, let's use `:` as separator in that case.

PR #230, opened 2022-12-05
Add the ability to specify the seed data path

RyuzakiKK:
When using the `--seed foo.caibx` option, Desync expects to find the data in foo, next to the provided .caibx file.

While this is a valid assumption for files, it requires additional steps when the data is a block device. For example a common suggestion in those cases is to create a symlink next to the seed index.

Instead, now `--seed` gains the ability to explicitly set a path for the data blob. The syntax becomes:
`--seed /path/to/index.caibx:/path/to/data`.

If the colon is missing, the old behavior of expecting the data next to the index file will still be used by default.

Fixes: #228


cmd/desync/extract.go
```diff
@@ -34,10 +35,12 @@ func newExtractCommand(ctx context.Context) *cobra.Command {
...
-Multiple optional seed indexes can be given with -seed. The matching blob needs
-to have the same name as the indexfile without the .caibx extension. If several
-seed files and indexes are available, the -seed-dir option can be used to
-automatically select call .caibx files in a directory as seeds. Use '-' to read
+Multiple optional seed indexes can be given with -seed. The matching blob should
+have the same name as the index file without the .caibx extension. Instead, if the
+matching blob data is in another location, or with a different name, you can explicitly
+set the path by writing the index file path, followed by a colon and the data path.
+If several seed files and indexes are available, the -seed-dir option can be used
+to automatically select call .caibx files in a directory as seeds. Use '-' to read
```
folbricht, line 43:
```suggestion
to automatically select all .caibx files in a directory as seeds. Use '-' to read
```
Old typo, but since you're changing it anyway :)

cmd/desync/extract.go
```diff
@@ -164,14 +168,33 @@ func writeInplace(ctx context.Context, name string, idx desync.Index, s desync.S
...
+			srcFile      string
+		)
+
+		if strings.HasSuffix(seedInfo, ".caibx") {
+			srcIndexFile = seedInfo
+			srcFile = strings.TrimSuffix(srcIndexFile, ".caibx")
+		} else {
+			seedArray := strings.SplitN(seedInfo, ":", 3)
+			if len(seedArray) < 2 {
+				return nil, fmt.Errorf("the provided seed argument \"%q\" seems to be malformed", seedInfo)
```
folbricht, line 185: `%q` already adds quotes in the formatted string, no need for `\"` as well

cmd/desync/extract.go
```diff
@@ -164,14 +168,33 @@ func writeInplace(ctx context.Context, name string, idx desync.Index, s desync.S
...
+		var (
+			srcIndexFile string
+			srcFile      string
+		)
+
+		if strings.HasSuffix(seedInfo, ".caibx") {
+			srcIndexFile = seedInfo
+			srcFile = strings.TrimSuffix(srcIndexFile, ".caibx")
+		} else {
+			seedArray := strings.SplitN(seedInfo, ":", 3)
```
folbricht, line 183: You don't really need to use SplitN here, a simple Split() would do as well which then removes the magic 3 and make it more clear

cmd/desync/extract.go
RyuzakiKK: Oh right, removed the quotes.

cmd/desync/extract.go
RyuzakiKK: Good point. I changed it to `Split()`.

folbricht (approved):
Thank you

PR #245, opened 2023-09-08
Improve options handling and add `--error-retry-base-interval`

RyuzakiKK:
- Remove deprecated "http-timeout" and "http-error-retry" options

  Those two options have been deprecated 5 years ago.
There is likely no reason to keep them around anymore.

- Keep the CLI options while querying the stores in info

- Correctly set the default option values and merge them with the config

  When loading a config file, we need to set the options to their default
values. Otherwise, when merging it with the CLI options, we could end up
with unexpected results.

  Additionally we need to distinguish between a CLI option that is set to
its default value because it was not provided as a launch parameter or
if it was explicitly set by the client. This is important because in
`MergedWith()` we should prefer the CLI option only if it was explicitly
set.

- Add --error-retry-base-interval as a CLI option

  Setting the base interval between error retries is a useful option,
especially when the Internet connection might be unstable.

  Given that `error-retry` can already be set via the CLI, by adding an
option for `error-retry-base-interval` too we can avoid the need to
force users to ship a config file when its only purpose was to only
change this value.

- Set a 500ms default value for error-retry-base-interval

  When the Internet connection is unstable, a few requests might fail. In
those cases, by default, Desync retries 3 times before giving up.

  However, if the connection momentarily becomes unstable and you retry
immediately without waiting, the chances of those attempts still failing
are very high.

  With this commit we set a more reasonable 500ms of base wait interval to
give the clients an higher chance of success.


RyuzakiKK:
> Remove deprecated "http-timeout" and "http-error-retry" options

This makes the parsing of the properties easier. AFAICT those has been marked as deprecated 5 years ago, so I just deleted them.
If you prefer to still keep them around, I can revisit this PR.

> Keep the CLI options while querying the stores in info

It's not entirely clear to me why the stores in `info` were queried with a default command option instead of using the user provided one. Please let me know in case I overlook something.

> Set a 500ms default value for error-retry-base-interval

If you prefer something different than 500ms, just let me know. The important thing for us is to have the `--error-retry-base-interval` so that we can set a reasonable wait interval without having to ship a config file just to set that value.

folbricht (approved):

folbricht:
Looks good. Thank you

PR #246, opened 2023-09-20
Add information about the cache

RyuzakiKK:
- Add information about the cache

  When using `desync info` it might be useful to have additional
information regarding your local cache.

  For example knowing the amount of chunks that can be picked up by your
cache can be used as a metric to ensure the cache is actually working as
expected.

  With this commit we add `in-cache` to print the number of chunks that
the cache already has, `not-in-seed-nor-cache` with the number of chunks
that needs to be downloaded from the store and finally
`dedup-size-not-in-seed-nor-cache` with the size of the chunks that
needs to be downloaded from the store.

- Reduce code duplication in info_test
- Add info test with cache and without a seed

### DIFF
--- a/cmd/desync/info.go
+++ b/cmd/desync/info.go
@@ -14,6 +14,7 @@ type infoOptions struct {
 	cmdStoreOptions
 	stores      []string
 	seeds       []string
+	cache       string
 	printFormat string
 }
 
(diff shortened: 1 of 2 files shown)


folbricht (approved):

folbricht:
Thank you

PR #257, opened 2024-02-16
Add examples for make and extract

bearrito:
Adding examples.

 I find there are many use cases that are covered in the source code docs that aren't ready to access examples. I added 2 of ones I had trouble locating. I also moved one existing example so it was closer to other similar examples.


folbricht (approved):
Thank you

Issue #262, opened 2024-03-08
Failing sparsefile tests


bearrito:
Sparsefile tests are flaky. There are some recent failing tests from CI with the same.

```
 sparse-file_test.go:103: 
        	Error Trace:	/home/bstrausser/Git/desync/sparse-file_test.go:91
        	Error:      	Received unexpected error:
        	            	EOF
        	Test:       	TestSparseFileRead
```

The error originates at

```
func (h *SparseFileHandle) ReadAt(b []byte, offset int64) (int, error) {
	if err := h.sf.loader.loadRange(offset, int64(len(b))); err != nil {
		return 0, err
	}
        // return errs
	return h.file.ReadAt(b, offset)
}

```

Which calls code that looks like
```
               // This is the offending syscall
		m, e := f.pread(b, off)
		if e != nil {
			err = f.wrapErr("read", e)
			break
		}
		n += m
		b = b[m:]
		off += int64(m)
	}
	return

```
In the above the off(set) will always be equal to 2097152 when it throws.

I believe the issue is the random choice of start and offset

The test code looks like. The commented values are the value of vars on a run that throws.
```
// 2097152
index_length := int(index.Length())
//2089125
start := rand.Intn(index_length)

//12025
length := rand.Intn(int(index.Index.ChunkSizeMax))

fromSparse := make([]byte, length)
fromBlob := make([]byte, length)

_, err := h.ReadAt(fromSparse, int64(start))
```
This will cause us to attempt to read 12,025 bytes starting at 2,089,125. 

There is a lot going on in creating a new sparse file, but I assume this is larger than the backing index file which has size 2,097,152

PR #263, opened 2024-03-09
Fix sparse file test offset/length

bearrito:
Fixes https://github.com/folbricht/desync/issues/262

### DIFF
--- a/sparse-file_test.go
+++ b/sparse-file_test.go
@@ -82,7 +82,7 @@ func TestSparseFileRead(t *testing.T) {
 	// Read a few random ranges and compare to the expected blob content
 	for i := 0; i < 1000; i++ {
 		length := rand.Intn(int(index.Index.ChunkSizeMax))
-		offset := rand.Intn(int(index.Length()) - length)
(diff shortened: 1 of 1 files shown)


sparse-file_test.go
```diff
@@ -79,18 +79,19 @@ func TestSparseFileRead(t *testing.T) {
 	require.NoError(t, err)
 	defer h.Close()
 
-	// Read a few randome ranges and compare to the expected blob content
-	for i := 0; i < 10; i++ {
-		start := rand.Intn(int(index.Length()))
+	// Read a few random ranges and compare to the expected blob content
+	for i := 0; i < 1000; i++ {
 		length := rand.Intn(int(index.Index.ChunkSizeMax))
+		offset := rand.Intn(int(index.Length()) - length)
```
folbricht, line 85: Shouldn't this be?
```go
offset := rand.Intn(int(index.Length()) - length - 1)
```

bearrito: Three hard things in computer science: off-by-one errors and naming things.

I'll fix it up.

folbricht (approved):

Issue #270, opened 2024-09-20
Add ability to calculate the exact size of the chunks necessary to be downloaded


RyuzakiKK:
## Current situation

We are using Desync to ship system updates, and in order to let the clients estimate the size of an update, we do the following:
- Download the `update.caibx`
- Generate a `current_rootfs.caibx` for the current active partition that will be used as a seed
  - Usually this step is skipped because we save the `.caibx` after applying an update
- Run something like `desync info --seed current_rootfs.caibx update.caibx`
- Parse the `dedup-size-not-in-seed` value
- Reduce that value by 33% to account for the compression rate

## Limitations of this solution

The problem is that `desync info` is only capable of providing information about the _uncompressed_ chunks size that you'll need to download. However, given that we are using compressed chunks, we have no way of knowing exactly how much data will be required to download.

The heuristic of a compression rate of 1.33 places us in the right ballpark, but the real compression rate can varies widely based on how easy/hard those particular chunks can be compressed.

Even if we could get a detailed list of all the chunks that need to be downloaded, it would be unfeasible to send thousands of requests to the server just to ask for their sizes.

### Option 1

One solution could be to extend the `.caibx`/`.caidx` files to include an additional (optional) field about the compressed size for each chunk. However I'm worried this might break backward compatibility with previous Desync versions, or with Casync.

Probably this would require some investigation to ensure we can preserve backward compatibility.

### Option 2

When running `desync make`, or with a separate command, we can store the compressed chunks size in another `.cai*` file. It would be some sort of external file to hold additional optional info about the main `.caibx`/`.caidx` file.

-----

@folbricht what do you think? Do you have other solutions in mind? Before starting spending time to actually design and implement something, I'd like to have your opinion first.

folbricht:
I'm not a fan of option 1. We shouldn't break compatibility for this. Especially since it's not going to be exact/reliable. That's due to stores potentially using different compression levels, a different version of the compression library, or no compression at all.

Do you need a precise number? If you're looking for a better estimate rather than a fixed compression ratio, you could use the chunks you have on the target already to calculate a better one (assuming the new chunks are compressed similarly).

As for option 2, you could achieve something similar today using a simple script. After running `make`, you can run `list` to get a full listing of all chunks in the index, then use that to lookup the file size on disk directly. You could have it produce a listing like so:

```
<chunkid1>,<compressed size1>
<chunkid2>,<compressed size2>
...
```

[...]

PR #273, opened 2024-10-28
Add more detailed information about the compressed chunks

RyuzakiKK:
`list-chunks` has now a new `--detailed` option that allows to print additional information about the chunks, including their compressed size (if a local store is provided).

One use case is for the clients to download the server's list-chunks JSON and feed it into `desync info`, to calculate the precise download size of an update.

Addresses https://github.com/folbricht/desync/issues/270


cmd/desync/list.go
```diff
@@ -41,15 +50,56 @@ func runList(ctx context.Context, opt listOptions, args []string) error {
...
+		if opt.detailed && opt.store != "" {
+			size, err = s.GetChunkSize(chunk.ID)
+			if err != nil {
+				// The compressed chunk is either missing or unreadable from the provided local store
+				size = 0
+			}
+		}
+
+		chunksInfo = append(chunksInfo, desync.ChunkAdditionalInfo{
+			ID:               chunk.ID.String(),
```
folbricht, line 81: Let's try to avoid using `.String()` so frequently. If you change the type of `ID` in `ChunkAdditionalInfo` to `ChunkID` this will integrate into the rest of the code base a lot better. You may need to add [custom Text marshaller/unmarshaller](https://pkg.go.dev/encoding/json#example-package-TextMarshalJSON) to the ChunkID type to have the ID as string in the JSON file.

local.go
```diff
@@ -240,9 +240,19 @@ func (s LocalStore) String() string {
 	return s.Base
 }
 
-// Close the store. NOP opertation, needed to implement Store interface.
+// Close the store. NOP operation, needed to implement Store interface.
 func (s LocalStore) Close() error { return nil }
 
+// GetChunkSize returns the bytes size of the chunk in this store
```
folbricht, line 246:
```suggestion
// GetChunkSize returns the bytes size of the raw, possibly compressed chunk in this store.
```

RyuzakiKK:
Thanks for the review. By the end of the week I should be able to find some time to address the points raised.

cmd/desync/list.go
RyuzakiKK: It should be done now.

cmd/desync/list.go
```diff
@@ -17,16 +20,22 @@ func newListCommand(ctx context.Context) *cobra.Command {
...
+desync list-chunks --store /mnt/store --detailed file.caibx`,
+		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runList(ctx, opt, args)
 		},
 		SilenceUsage: true,
 	}
 	flags := cmd.Flags()
+	flags.BoolVarP(&opt.detailed, "detailed", "d", false, "detailed chunk info")
+	flags.StringVarP(&opt.store, "store", "s", "", "local source store")
```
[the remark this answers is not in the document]
RyuzakiKK: Yeah, makes sense. I created `inspect-chunks` because I kept the local store as optional. If you don't provide a local store, the JSON will only contain the list of chunks hashes and uncompressed sizes.

Probably not something that is terribly useful to have, but I tired to keep this new command as generic as possible. In this way, if in the future we want to add other elements as well, it should be easier to expand.

Of course I'm open to suggestions here.

cmd/desync/list.go
```diff
@@ -41,15 +50,56 @@ func runList(ctx context.Context, opt listOptions, args []string) error {
...
+
+	var s desync.LocalStore
+	if opt.store != "" {
+		options, err := cfg.GetStoreOptionsFor(opt.store)
+		if err != nil {
+			return err
+		}
+		// Use compressed chunks, if available, to record their sizes
+		options.Uncompressed = false
+		s, err = desync.NewLocalStore(opt.store, options)
```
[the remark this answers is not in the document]
RyuzakiKK: The new command explicitly mention that we need a local store. So it should be better now.

cmd/desync/inspectchunks_test.go
```diff
@@ -0,0 +1,57 @@
...
+		{"inspect the chunks info without any stores",
+			[]string{"testdata/blob2.caibx"},
+			"testdata/blob2_chunks_info_no_store.json",
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			var exp []desync.ChunkAdditionalInfo
+			be, err := os.ReadFile(test.expectedOutputJSON)
+			require.NoError(t, err)
+			err = json.Unmarshal(be, &exp)
```
folbricht, line 38: Missing an error-check here?

cmd/desync/info.go
```diff
@@ -27,10 +30,13 @@ func newInfoCommand(ctx context.Context) *cobra.Command {
 		Long: `Displays information about the provided index, such as the number of chunks
 and the total size of unique chunks that are not available in the seed. If a
 store is provided, it'll also show how many of the chunks are present in the
-store. If one or more seed indexes are provided, the number of chunks available
+store. By providing a chunks info file, generated by 'inspect-local', additional
```
folbricht, line 33:
```suggestion
store. By providing a chunks info file, generated by 'inspect-chunks', additional
```

cmd/desync/inspectchunks.go
```diff
@@ -0,0 +1,101 @@
...
+
+	var s desync.LocalStore
+	if opt.store != "" {
+		options, err := cfg.GetStoreOptionsFor(opt.store)
+		if err != nil {
+			return err
+		}
+
+		// Use compressed chunks, if available, to record their sizes
+		options.Uncompressed = false
```
folbricht, line 74: This doesn't look quite right and is confusing for the user. We should allow the user to specify if the store is compressed or not and not override configuration. Wouldn't it be better to simply skip `size, _ = s.GetChunkSize(chunk.ID)` in the loop below when `options.Uncompressed` is true?

cmd/desync/inspectchunks.go
```diff
@@ -0,0 +1,101 @@
...
+	c, err := readCaibxFile(args[0], opt.cmdStoreOptions)
+	if err != nil {
+		return err
+	}
+
+	var chunksInfo []desync.ChunkAdditionalInfo
+
+	var s desync.LocalStore
+	if opt.store != "" {
+		options, err := cfg.GetStoreOptionsFor(opt.store)
```
folbricht, line 68: To improve the UX of this, would you mind initializing the store like in https://github.com/folbricht/desync/blob/master/cmd/desync/prune.go#L50-L58 ?
- Use the existing function to initialize the store
- Then confirm it's a `LocalStore`
- Error if it's something else that's not supported
This should align better with the other commands and provide a clearer error message.

folbricht (commented):
It's close, just left some minor comments

cmd/desync/inspectchunks.go
RyuzakiKK: Yeap, now I don't override that option and I added `!s.Opt.Uncompressed` to the `if`.

folbricht (approved):
Thank you. Would it be worth adding an example to the README for this?

PR #274, opened 2024-11-27
Fix file mode preservation during untar if mtime=0

matshch:
This commit fixes file mode preservation after `desync tar --no-time` and `desync untar`.


folbricht (approved):
Thank you

PR #283, opened 2025-03-30
Allow `chunk`/`make` to accept `-` as a shorthand for reading from `stdin`

staticfloat:
This extends the `chunk` and `make` commands to be able to read from stdin when `-` is passed as a filename, as is UNIX tradition.  Note that this implementation is purposefully quite basic and straightforward; when reading from stdin all concurrency is disabled for `make`, and `catar` input is not handled the same way due to the need to parse the header first.  For my purposes (which always deals with blobs rather than `catar` inputs, this is sufficient, however if there's a simple way to allow peeking the input stream then feeding it in to be hashed again, I'm happy to include it here.

### DIFF
--- a/cmd/desync/chunk.go
+++ b/cmd/desync/chunk.go
@@ -43,12 +43,17 @@ func runChunk(ctx context.Context, opt chunkOptions, args []string) error {
 
 	dataFile := args[0]
 
-	// Open the blob
-	f, err := os.Open(dataFile)
-	if err != nil {
-		return err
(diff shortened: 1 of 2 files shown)


staticfloat:
Closing this because my implementation was incomplete, and fundamentally flawed, as I now see that `make` performs two passes over the input if a chunk store is passed.  While I could try to have the indexing also write out chunks to the store in a single pass, that's more than I want to attempt here.

PR #307, opened 2026-02-08
Optimize chunker hot loop for ~1.5-2x throughput

### LINKED ISSUE #244: borg is faster than "desync make" despite borg is single-threaded
Description. borg2 is significantly (x1.2-x2) faster than desync when storing particular 10 GiB file to empty storage despite borg2 is single-threaded and desync is multi-threaded. I tested this on two different machines. And both time I get significant time difference. (But theoretically desync should be significantly faster than borg, because desync is parallel). Same compression and chunking settings was used.

Steps to reproduce.

Install borg. Make sure you installed borg2 as opposed to borg1. (I tested borg2 only, I don't know how well borg1 performs). I used borg 2.0.0b5 for my tests. I installed borg2 from debian repository.

Build desync. I used desync from a0052781292e3d140f6f117a2d608a22471e86e8. I installed debian package `golang-go` and did `git clone https://github.com/folbricht/desync.git; cd desync/cmd/desync; go build; cp desync /`.

Make sure all data (both input file and repositories) located on tmpfs.

Also make sure you always add data (using `desync make` or `borg create`) to empty repository.

Grab this file: http://safinaskar.com/f/03.zst and decompress it. This is zstd-compressed VM image. The image contains nothing confidential, so can be distributed without restriction. This is simple Debian VM. Its size is 490M. After decompression you will get 10 GiB file.

Make sure that after decompression the file is not sparse. If you are unsure, then copy it using `cp --sparse=never`.

Create desync repo using these commands:
```bash
mkdir /home/user/dedup-bench/input-fs/desync-repo
mkdir /home/user/dedup-bench/input-fs/desync-repo/index
mkdir /home/user/dedup-bench/input-fs/desync-repo/storage.castr
```
(Of course, exact path doesn't matter. I copied here my actual path, but you can use some other path.)

Put `03` to desync repo using this command:
```bash
time -p /desync make -m 16:64:256 -s /home/user/dedup-bench/input-fs/desync-repo/storage.castr /home/user/dedup-bench/input-fs/desync-repo/index/a.caibx  /home/user/dedup-bench/input-fs/03
```

Create borg repo using these commands:
```bash
export BORG_PASSPHRASE=password
borg2 rcreate --encryption=authenticated-blake2 --repo=/home/user/dedup-bench/input-fs/borg-repo
```

That directory (`/home/user/dedup-bench/input-fs/borg-repo` in my case) should not exist before this command (i. e. `borg2 rcreate`).

Add `03` to borg repo using these commands:
```bash
export BORG_PASSPHRASE=password
time -p borg2 create --repo=/home/user/dedup-bench/input-fs/borg-repo --chunker-params buzhash,14,18,16,4095 --compression zstd,3 a /home/user/dedup-bench/input-fs/03
```

End of steps to reproduce.

In both cases we used same settings. zstd level 3 (as well as I understand, in desync we have zstd level 3 hardcoded). And buzhash (again, as well as I understand desync uses buzhash).

I used string `16:64:256` for configuring desync. As well as I understand this means that desync should use 16 KiB as minimal chunk size, 64 KiB as average and 256 KiB as maximal. I used string `buzhash,14,18,16,4095` for configuring borg. This means using `2 ^ 14` bytes as minimal chunk size, `2 ^ 16` bytes as average and `2 ^ 18` bytes as maximal. I. e. the same.

I did this test on two machines. First machine is AWS bare metal server. It was configured for benchmarks using these advises: https://easyperf.net/blog/2019/08/02/Perf-measurement-environment-on-Linux . It has 48 cpu cores, so I resticted my tests to 4 cpu cores by prepending `taskset 0x0000000F` to commands. (You may say that `taskset` confused desync and desync was not able to discover available parallelism. I don't think so, because test results on this machine match test results on another machine, where taskset was not used.) (Side note: if I don't restrict execution to 4 cpu cores and use all 48 cpu cores, then desync becomes faster than borg.)

AWS is debian bookworm.

Okay, so here are AWS results:
```text
admin@ip-172-31-23-30:~$ time -p taskset 0x0000000F /desync make -m 16:64:256 -s /tmp/tmpfs/desync-repo/storage.castr /tmp/tmpfs/desync-repo/index/a.caibx /tmp/tmpfs/sparse-never-sto/03
Chunking [=========================================================================================================================================] 100.00% 20s
Storing [===========================================================================================================================================] 100.00% 9s
real 29.58
user 110.99
sys 4.42

admin@ip-172-31-23-30:/tmp/tmpfs$ BORG_PASSPHRASE=password time -p taskset 0x0000000F borg2 create --repo=/tmp/tmpfs/repo --chunker-params buzhash,14,18,16,4095 --compression zstd,3 a /tmp/tmpfs/sparse-never-sto/03 
real 24.39
user 22.65
sys 1.81
```

As you can see borg time is 24.39 and desync time is 29.58, i. e. desync is x1.2 slower.

Second machine is my personal laptop Dell Inspiron. Tests was performed in debian sid inside of docker container in debian stretch. Machine has 4 cpus. No `taskset` was used. Machine was not tuned for benchmarks. It was under heavy load. But I did tests two times and both times difference was huge. So I assume heavy load was not reason for such results. Okay, so here are results:
```text
<[sid]>root@4af97530f96d:~# time -p /desync make -m 16:64:256 -s /home/user/dedup-bench/input-fs/desync-repo/storage.castr /home/user/dedup-bench/input-fs/desync-repo/index/a.caibx  /home/user/dedup-bench/input-fs/03
Chunking [=========================================================================================================================================] 100.00% 26s
Storing [==========================================================================================================================================] 100.00% 15s
real 42.22
user 136.80
sys 5.52

<[sid]>root@4af97530f96d:~# time -p /desync make -m 16:64:256 -s /home/user/dedup-bench/input-fs/desync-repo/storage.castr /home/user/dedup-bench/input-fs/desync-repo/index/a.caibx  /home/user/dedup-bench/input-fs/03
Chunking [=========================================================================================================================================] 100.00% 28s
Storing [==========================================================================================================================================] 100.00% 15s
real 43.42
user 141.85
sys 5.13

<[sid]>root@29ffe8165cb8:/home/user/dedup-bench/input-fs# BORG_PASSPHRASE=password time -p borg2 create --repo=/home/user/dedup-bench/input-fs/borg-repo --chunker-params buzhash,14,18,16,4095 --compression zstd,3 a /home/user/dedup-bench/input-fs/03
real 23.74
user 19.69
sys 2.19

<[sid]>root@29ffe8165cb8:/home/user/dedup-bench/input-fs# BORG_PASSPHRASE=password time -p borg2 create --repo=/home/user/dedup-bench/input-fs/borg-repo --chunker-params buzhash,14,18,16,4095 --compression zstd,3 a /home/user/dedup-bench/input-fs/03
real 20.82
user 19.46
sys 1.30
```
As you can see, desync is nearly x2 slower than borg. Repo size for desync is 525M and repo size for borg is 522M.

List of borg's optimizations:
- borg uses blake2b as its strong hash for comparing chunks (in this test)
- borg has special support for storing sparse files to its repo (but this should not matter for this test, because our file is not sparse) (this optimization is present in `fixed` codepath, I don't know whether it is present in `buzhash` codepath)
- borg checks whether given chunk is all-zero. If it is and hash (i. e. blake2b for this test) for chunk of such size was already computed before, then borg takes already computed hash (this optimization is present in `fixed` codepath, I don't know whether it is present in `buzhash` codepath) (borg does this optimization even if file is not sparse)
- after reading chunk borg drops it from OS cache using posix_fadvise with DONTNEED (I saw this optimization in `fixed` codepath). But this should not matter in our case, because in our case the file is stored in tmpfs

folbricht:
## Summary

- Replace expensive `DIVL` instruction in boundary detection (`hValue % disc == disc-1`) with **Lemire's fast divisibility test** — a multiply-and-compare that costs ~5 cycles vs ~26 for hardware division
- Convert `hashTable` from slice to `[256]uint32` fixed-size array, eliminating bounds checks on all accesses
- Add precomputed `hashTableRotated` table, removing one `RotateLeft32` call per byte in the hot loop
- **Hoist struct fields** (`hValue`, `hIdx`, `buf`, window pointer, Lemire constants) into local variables to avoid repeated pointer dereferences
- Replace `(hIdx + 1) % 48` with a branch (`if hIdx >= 48`) — perfectly predicted, taken once per 48 iterations
- **Reuse backing buffer** in `fillBuffer()` to reduce allocations from O(filesize/buffersize) to O(1)
- Add `b.SetBytes()` to benchmarks for direct MB/s throughput reporting; modernize to use `b.Loop()`

All optimizations preserve **identical chunk boundaries**, verified by `TestChunkerLargeFile` which checks exact SHA512/256 hashes for every chunk.

## Test plan

- [x] `go test -run TestChunkerLargeFile` — exact chunk boundary regression (SHA512/256 hashes)
- [x] `go test -run TestChunker` — all chunker tests pass
- [ ] `go test -bench=BenchmarkChunk -benchmem` — verify throughput improvement
- [ ] CI passes on Linux, macOS, Windows

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Before:
```
goos: linux
goarch: amd64
pkg: github.com/folbricht/desync
cpu: Intel(R) Core(TM) i7-7600U CPU @ 2.80GHz
BenchmarkChunker-4         	     272	   4304447 ns/op	 2621489 B/op	       2 allocs/op
BenchmarkChunkNull1M-4     	     205	   5907772 ns/op	 2621489 B/op	       2 allocs/op
BenchmarkChunkNull10M-4    	      15	  69146171 ns/op	13107254 B/op	       6 allocs/op
BenchmarkChunkNull50M-4    	       4	 294544244 ns/op	55050312 B/op	      23 allocs/op
BenchmarkChunkNull100M-4   	       2	 593006637 ns/op	107479112 B/op	      43 allocs/op
```

After:
```
goos: linux
goarch: amd64
pkg: github.com/folbricht/desync
cpu: Intel(R) Core(TM) i7-7600U CPU @ 2.80GHz
BenchmarkChunker-4         	     472	   2371186 ns/op	 442.22 MB/s	 2621489 B/op	       2 allocs/op
BenchmarkChunkNull1M-4     	     376	   3125929 ns/op	 335.44 MB/s	 2621490 B/op	       2 allocs/op
BenchmarkChunkNull10M-4    	      38	  31382746 ns/op	 334.12 MB/s	 2621489 B/op	       2 allocs/op
BenchmarkChunkNull50M-4    	       7	 153479795 ns/op	 341.60 MB/s	 2621488 B/op	       2 allocs/op
BenchmarkChunkNull100M-4   	       4	 307347136 ns/op	 341.17 MB/s	 2621488 B/op	       2 allocs/op
```

Closes https://github.com/folbricht/desync/issues/244

### DIFF
--- a/chunker.go
+++ b/chunker.go
@@ -14,7 +14,20 @@ func discriminatorFromAvg(avg uint64) uint32 {
 	return uint32(float64(avg) / (-1.42888852e-7*float64(avg) + 1.33237515))
 }
 
-var hashTable = []uint32{
+// modInverse32 computes the modular multiplicative inverse of an odd number d
+// modulo 2^32 using Newton's method. This is used for Lemire's fast
+// divisibility test which replaces expensive hardware division.
+func modInverse32(d uint32) uint32 {
+	x := d             // d is odd, so d*d ≡ 1 (mod 4) is a valid start
+	x *= 2 - d*x       // 3 bits
+	x *= 2 - d*x       // 6 bits
+	x *= 2 - d*x       // 12 bits
+	x *= 2 - d*x       // 24 bits
+	x *= 2 - d*x       // 48 bits → full 32-bit precision
+	return x
+}
+
+var hashTable = [256]uint32{
 	0x458be752, 0xc10748cc, 0xfbbcdbb8, 0x6ded5b68,
 	0xb10a82b5, 0x20d75648, 0xdfc5665f, 0xa8428801,
 	0x7ebf5191, 0x841135c7, 0x65cc53b3, 0x280a597c,
@@ -81,6 +94,16 @@ var hashTable = []uint32{
 	0x7bf7cabc, 0xf9c18d66, 0x593ade65, 0xd95ddf11,
 }
 
+// hashTableRotated contains hashTable values pre-rotated by ChunkerWindowSize,
+// eliminating a RotateLeft32 call per byte in the hot loop.
+var hashTableRotated [256]uint32
+
+func init() {
+	for i := range hashTable {
+		hashTableRotated[i] = bits.RotateLeft32(hashTable[i], ChunkerWindowSize)
+	}
+}
+
 // Chunker is used to break up a data stream into chunks of data.
 type Chunker struct {
 	r             io.Reader
@@ -88,14 +111,25 @@ type Chunker struct {
 
 	start uint64
 
-	buf    []byte
-	hitEOF bool // true once the reader returned EOF
+	buf        []byte
+	backingBuf []byte // reusable backing buffer for fillBuffer
+	hitEOF     bool   // true once the reader returned EOF
 
 	// rolling hash values
 	hValue         uint32
 	hWindow        [ChunkerWindowSize]byte
 	hIdx           int
 	hDiscriminator uint32
+
+	// Precomputed values for Lemire's fast divisibility test.
+	// Tests "hValue % d == d-1" without hardware division by rewriting as
+	// "((hValue - (d-1)) >> trailShift) * inverseOdd < threshOdd"
+	// where d's power-of-2 factor is separated out.
+	hInverseOdd uint32 // modular inverse of odd part of hDiscriminator
+	hThreshOdd  uint32 // ceil(2^32 / odd part of hDiscriminator)
+	hDiscMinus1 uint32 // hDiscriminator - 1
+	hTrailShift uint   // number of trailing zeros in hDiscriminator
+	hTrailMask  uint32 // (1 << hTrailShift) - 1, for checking low bits
 }
 
 // NewChunker initializes a chunker for a data stream according to min/avg/max chunk size.
@@ -112,12 +146,29 @@ func NewChunker(r io.Reader, min, avg, max uint64) (Chunker, error) {
 	if avg > max {
 		return Chunker{}, errors.New("avg chunk size must not be greater than max")
 	}
+	disc := discriminatorFromAvg(avg)
+
+	// Precompute Lemire's fast divisibility constants.
+	// We want to test: hValue % disc == disc-1
+	// Rewritten as: (hValue - (disc-1)) is divisible by disc
+	// Factor disc = odd * 2^shift, then use modular inverse of odd part.
+	trailShift := uint(bits.TrailingZeros32(disc))
+	oddPart := disc >> trailShift
+	inverseOdd := modInverse32(oddPart)
+	// threshOdd = ceil(2^32 / oddPart)
+	threshOdd := uint32(((1 << 32) + uint64(oddPart) - 1) / uint64(oddPart))
+
 	return Chunker{
 		r:              r,
 		min:            min,
 		avg:            avg,
 		max:            max,
-		hDiscriminator: discriminatorFromAvg(avg),
+		hDiscriminator: disc,
+		hInverseOdd:    inverseOdd,
+		hThreshOdd:     threshOdd,
+		hDiscMinus1:    disc - 1,
+		hTrailShift:    trailShift,
+		hTrailMask:     (1 << trailShift) - 1,
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


matshch:
We see huge improvements on our workloads (even slightly more than 2x), thanks a lot!

folbricht:
Glad it's working well for your use-case.

folbricht:
@matshch There's now https://github.com/folbricht/desync/pull/367 which should speed things up even further. Would you like to try it out before I merge?

matshch:
@folbricht, thanks for sharing that! I'll try to check it out in the upcoming days.

PR #379, opened 2026-08-03
Use value receivers for `ChunkID` `String` and `MarshalJSON`

matshch:
`ChunkID.String()` was switched to a pointer receiver in #273, so only `*ChunkID` implemented `fmt.Stringer`. A `ChunkID` value passed to `%s` or `%v` was rendered by `fmt`'s byte-array fallback, printing the 32 raw bytes instead of hex. This affected error and log messages in `s3.go`, `gcs.go`, `httphandler.go`, and `cmd/desync/info.go`, plus test failure messages in `s3_test.go`.

Switching `String()` to a value receiver puts it in the method set of both `ChunkID` and `*ChunkID`, fixing all format call sites at once. The same change to `MarshalJSON` removes the addressability trap where `json.Marshal` of a non-addressable `ChunkID` would emit a number array instead of a hex string. `UnmarshalJSON` keeps its pointer receiver as it mutates the receiver.

Also drop explicit `.String()` calls in `fmt` and `logrus` arguments where the `Stringer` is now picked up automatically; calls that produce an actual string value (path building, string-typed parameters) remain.

### DIFF
--- a/assemble.go
+++ b/assemble.go
@@ -73,7 +73,7 @@ func writeChunk(c IndexChunk, ss *selfSeed, f *os.File, blocksize uint64, s Stor
 	}
 	// Might as well verify the chunk size while we're at it
 	if c.Size != uint64(len(b)) {
-		return fmt.Errorf("unexpected size for chunk %s", c.ID.String())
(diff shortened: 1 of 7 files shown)


folbricht:
Thank you

Issue #225, opened 2022-10-18
Add man page


dvzrv:
Hi! I'm packaging desync for Arch Linux.

I noticed that this project does not offer a man page for the `desync` executable.
This would be very helpful to have, so that people do not have to call the actual executable to read about its options.
Additionally, we (and other distributions, too) have a man page indexing service, which allows to read the man pages of packages online: https://man.archlinux.org
The man page for `desync` would automatically end up there and can serve as a reference point for further documentation (e.g. in our wiki or in other places).

folbricht:
Great, that would be a good addition to have for sure. I haven't written man pages before but initially we could use the content of the various help texts. Is there a certain layout/structure that would help you package them? I wouldn't mind if you made a PR just for the outline of what it should look like and I fill in the content.

folbricht:
https://github.com/folbricht/desync/pull/232 adds a new 'manpage' command which lets you generate them. Does that work for your build-system? Or would you prefer I generate a set and commit into the code?

folbricht:
Man pages now ship with releases, as of #399.

- Every release archive contains a `man/` directory with a page per command, generated at build time from the command definitions.
- They're now written to **section 1** (user commands). They had been going to section 3 (library calls), which would have been wrong for the `man1/` install path discussed above — worth knowing if you ever generated them with the `manpage` command before this.
- `desync manpage <dir>` still generates them on demand, and now creates the output directory rather than failing if it doesn't exist.
- Argument placeholders survive into the pages. They were previously dropped by the markdown-to-roff conversion, so a synopsis read `desync chop [flags]` and prose lost the argument mid-sentence.

Closing as addressed: packaging can install the pages straight out of the release tarball. There's still no Makefile with `DESTDIR`/`PREFIX` install targets as suggested in the thread — if that's still useful for packaging rather than just convenient, please open a separate issue for it and it can be looked at on its own.

Thanks for the original report, and apologies it sat for so long.

PR #399, opened 2026-08-30
Generate the CLI reference and ship the man pages

folbricht:
The README documented every flag by hand, and had drifted:

- **27 of the 62 flags** the commands define were never mentioned anywhere on the page — `--encryption-algorithm`, `--ignore-chunks`, `--one-file-system`, `--skip-verify-write`, `--tar-add-root`, the whole `--cor-*` group, and others.
- **`--writeable` appeared twice** for a flag actually spelled `--writable` — once in a table, once in an `index-server` line meant to be copied and pasted. That one produces `unknown flag`.

## Generated reference

A hidden `gendocs` command writes a markdown page per command into [`docs/cli`](docs/cli), and CI regenerates and diffs it. The reference can now only be wrong by being stale, and CI catches that. Generation clears the directory first, otherwise the page for a deleted command would survive regeneration and the diff would have nothing to catch. `DisableAutoGenTag` is set so pages don't carry the date they were written, which would fail the check a day after any change.

It's hidden because it maintains this repository rather than doing anything for someone using the tool.

The README keeps its option tables — they explain the options rather than enumerating them, which the generated help text can't do — and now points at the generated reference for full flag lists. A second CI check verifies every flag the README mentions actually exists, which is the part generation can't cover. It fails on the `--writeable` typo and passes once fixed; both directions were tested under `bash -e`, as Actions runs it.

## Man pages (#225)

The `manpage` command has existed for a while but releases only ever shipped the binary. They're now generated in the goreleaser `before` hook and included in the archives — a build artifact, so not committed. Verified against a real snapshot build: 28 pages land in every archive alongside `LICENSE` and `README.md`.

Two fixes were needed on the way, both of which only mattered once the pages were actually shipped:

- They were written to **section 3** (library calls) rather than **section 1** (user commands), so an installed page went where `man desync` wouldn't look.
- Neither `manpage` nor `gendocs` created the output directory, which cobra expects to already exist. `desync manpage /tmp/man` failed unless you had made the directory yourself.

The `--source` default is now `desync` rather than leaving cobra to stamp "Auto generated by spf13/cobra" across the header of every shipped page.

## Argument placeholders

md2man parses `<index>` as an inline HTML tag and drops it, which cost more than the synopsis. Prose naming an argument lost it mid-sentence, so `desync-chop.1` shipped as:

> use `--ignore`  which will skip any chunks from the given index. The same can be achieved by providing the chunks in their ASCII representation in a text file with `--ignore-chunks` .

Escaping the angle bracket for the duration of generation makes the markdown parser emit it as literal text; the descriptions are restored afterwards, since the command tree is shared with the running CLI and the escape would otherwise reach `--help`. No `Use` string changes, so help output is untouched.

Two things worth knowing about the escape, both found by checking the generated pages rather than assuming:

- **It can't be applied to every angle bracket.** The completion command's help contains `source <(desync completion bash)`, and inside a code block the backslash survives into the page — rendering a command that fails if copied. Only placeholder-shaped tokens are escaped.
- **The pattern needs alternation.** `mtree` takes `<catar|index|dir>`, which a word-characters-only pattern misses, leaving that one page broken while the other 27 looked fixed.

Two tests cover it, both verified to fail with the fix disabled: one that the synopsis, the flag prose and the alternation survive with no escape reaching any page, one that `Use` and `Long` are restored after generation. All 28 pages were checked through `groff`.


folbricht:
Fixed the placeholder gap noted above rather than leaving it, in `12cf626`.

The damage went past the synopsis. Prose naming an argument lost it mid-sentence too, so `desync-chop.1` read:

> use `--ignore`  which will skip any chunks from the given index. The same can be achieved by providing the chunks in their ASCII representation in a text file with `--ignore-chunks` .

Escaping the angle bracket for the duration of generation makes the markdown parser emit it as literal text. The descriptions are restored afterwards, since the command tree is shared with the running CLI and the escape would otherwise show up in `--help`.

Two things the first attempt got wrong, both caught by checking the output rather than assuming:

- **Escaping every angle bracket broke a code block.** The completion command's help contains `source <(desync completion bash)`; inside a code block the backslash survives into the page, so it rendered as `source \<(desync completion bash)` — a command that no longer works if copied. Only placeholder-shaped tokens are escaped now.
- **The pattern needed alternation.** `mtree` takes `<catar|index|dir>`, which a word-characters-only pattern doesn't match, so that one page kept its broken synopsis while every other page looked fixed.

Two tests cover it: one asserting the synopsis, the flag prose and the alternation survive, with no escape reaching any page; one asserting `Use` and `Long` are restored after generation. Both fail without the fix — I checked, rather than trusting a passing test.

All 28 pages verified through `groff`.

PR #415, opened 2026-09-02
Skip Setgid test when we cannot set this bit

matshch:
Some environments are too restricted to set setgid, so check that we can do that before expecting the rest of the code to do that.

### DIFF
--- a/localfs_dirmeta_test.go
+++ b/localfs_dirmeta_test.go
@@ -109,6 +109,19 @@ func TestLocalFSDirModeWhilePopulated(t *testing.T) {
 // TestLocalFSDirSetgidWhilePopulated confirms the setgid bit is in place while
 // the contents of a directory are written, so they inherit its group.
 func TestLocalFSDirSetgidWhilePopulated(t *testing.T) {
+	// Some build sandboxes prevent us from setting setgid bit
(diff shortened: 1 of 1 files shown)


folbricht:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

The probe mirrors the sequence the code under test actually uses — `prepareDirWrite` creates the directory with `n.Mode.Perm()` (setgid stripped) and chmods the bit back in — so `os.Mkdir` followed by `os.Chmod(…|os.ModeSetgid)` is a faithful capability check, and the stat re-read covers the case where chmod reports success but the bit is not preserved.

https://github.com/folbricht/desync/blob/f1a54c4611088bb94d5a2d0fc224c949857ab786/localfs_other.go#L106-L114

The skip is additive: the assertions are untouched, so coverage is unchanged wherever setgid works. It also matches the existing convention for environment-dependent skips in this package.

https://github.com/folbricht/desync/blob/7092cc707a04017073eafa5af267857052f05b1a/localfs_dirmeta_test.go#L110-L124

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

folbricht (approved):

